PR #1, opened 2025-09-29
Explicit display_last4 length

mjallday:
Force display_last4 to be exactly four digits in length.


mjallday:
PR number 1 FTW! 🙌🏽

PR #3, opened 2025-09-30
Support decimal quantities for B2B commerce use cases

AlexandrePh:
Changed the quantity field from integer to number (decimal) to support B2B commerce scenarios where items are sold by weight (e.g., kilograms) or other fractional units of measurement.

Changes:

- Updated JSON schema to use type: number with exclusiveMinimum: true

- Updated OpenAPI spec to reflect decimal quantity support

- Updated all examples to use 2.5 as demonstration value

- Added changelog entry documenting this breaking change


spec/json-schema/schema.agentic_checkout.json
```diff
@@ -66,8 +66,9 @@
           "type": "string"
         },
         "quantity": {
-          "type": "integer",
-          "minimum": 1
+          "type": "number",
```
andreas-wickberg-vaimo, line 69: How about including a unit? Eg, a product sold by meter with decimal; a us shopper wanting to buy 8 feet 4 inches.

prasad-stripe (approved):
Makes sense to change the quantity to a number

PR #11, opened 2025-10-01
Fix FulfillmentOption money fields type inconsistency

MariuszT:
Hi,

I noticed that the spec has `subtotal`, `tax`, and `total` defined as `string` in both FulfillmentOption types, but the JSON Schema has them as `integer`. The examples also use numbers, not strings. Looking at the rest of the spec, all other money fields (LineItem, Total) use `integer` for minor units, so this looks like a typo.

Let me know if I missed something or if there was a reason for using strings here.

Thanks!

### DIFF
--- a/spec/openapi/openapi.agentic_checkout.yaml
+++ b/spec/openapi/openapi.agentic_checkout.yaml
@@ -377,9 +377,9 @@ components:
         carrier: { type: string }
         earliest_delivery_time: { type: string, format: date-time }
         latest_delivery_time: { type: string, format: date-time }
-        subtotal: { type: string }
-        tax: { type: string }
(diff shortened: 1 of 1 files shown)


prasad-stripe:
Thanks @MariuszT for sending this in — fixing FulfillmentOption so subtotal, tax, and total are integers in the OpenAPI spec is the right call and matches our “amounts in minor units” rule.
The repo has since moved to versioned specs (spec/2025-09-29/, etc.), and in those versions those fields are already defined as integers (and in newer versions we use a totals array for fulfillment options). So this change is effectively already reflected on main and there’s nothing left to merge from this branch.
We really appreciate you taking the time to improve the spec.

Issue #25, opened 2025-10-14
Question: Product Feed Spec


oslook:
Although you can see the definition of Product Feed on the following page,
https://developers.openai.com/commerce/specs/feed

there is no complete YAML or OpenAPI in the repo. Please confirm whether it needs to be added or clarified.

pauldraper:
> Deliver the feed. Share the feed using the preferred delivery method and file format described in the integration section.

> Integration guidance. The spec defines the preferred delivery method and file format for reliable ingestion.

The spec does not in fact define the preferred delivery method (SFTP, HTTPS, etc).

prasad-stripe:
We are still working on the standardizing the product feed. Suggestions are welcome! Or stay tuned, we are working on it.

PR #36, opened 2025-11-02
Adds product feed spec

ChrisLaRocque:
Apologies if this is already in progress, but noticed the product feed missing from this repository yet present in the [docs](https://developers.openai.com/commerce/specs/feed/) (as did others: #25 ).

There are some types that are not clear in the docs, I took my best guess but they should be reviewed. Some examples:

- Any type classified as a combination of 2 types, for example `pickup_sla` is listed as "Number + duration". For the time being I've classified them as strings.
- `shipping` mentions there can be several values separated by colons, but I'm unsure if that is true given the colons being used as delimiters in the example, or if they should in fact be separated by _commas_
- `product_category` mentions it should be of type "Category taxonomy" but no reference/examples are provided, so I typed as string for now.


prasad-stripe:
Thanks for this PR — we really like the direction.
No need for Apologies. The product feed spec is exactly the kind of thing we care about: structured catalog data so agents can discover and surface products accurately. Your RFC and schema are clear and well thought out.
We’re not quite ready to merge this yet — We’ll need a rebase onto current main and restructured it into a SEP, when we’re ready to integrate. In the meantime, hold tight: product discovery and feeds are an important area we want to invest in soon, and we’ll come back to this and work with you to get it in. We’ll follow up when we’re ready to move forward.
Thanks again for contributing.

PR #54, opened 2025-12-17
[SEP] Intent Traces: Structured cancellation context for checkout sessions

0xtotaylor:
## Abstract

This RFC extends the `POST /checkout_sessions/{id}/cancel` endpoint to accept an optional `intent_trace` object, enabling agents to transmit structured data about why a checkout session was abandoned.

## Motivation

In the current specification, cancellation is a terminal action with no structured context. Merchants receive a `status: canceled` session, but no signal about the user's objection. This forces reliance on probabilistic retargeting—buying ad impressions without knowing why the user left.

### Why structured cancellation context?

When an agent manages checkout, abandonment is often a reasoned decision based on specific constraints (price, shipping speed, trust concerns). Capturing this decision as an **Intent Trace** converts abandonment from a dead end into an actionable signal. A merchant receiving `reason_code: shipping_cost` can programmatically respond with a free shipping offer rather than guessing.

### Why optional?

The `intent_trace` is optional to maintain backward compatibility. Agents that do not support this extension omit the request body, and cancellation proceeds as before. Servers that do not implement it ignore the body and return `200 OK` with `status: canceled`.

### Endpoint Update: Cancel Session

**`POST /checkout_sessions/{checkout_session_id}/cancel`**

Extend the cancel endpoint to accept an optional request body with `Content-Type: application/json`.

```yaml
requestBody:
  required: false
  content:
    application/json:
      schema:
        $ref: "#/components/schemas/CancelSessionRequest"
```

### New Schema: CancelSessionRequest

```yaml
CancelSessionRequest:
  type: object
  additionalProperties: false
  properties:
    intent_trace:
      $ref: "#/components/schemas/IntentTrace"
```

### New Schema: IntentTrace

```yaml
IntentTrace:
  type: object
  additionalProperties: false
  properties:
    reason_code:
      type: string
      enum:
        - price_sensitivity
        - shipping_cost
        - shipping_speed
        - product_fit
        - trust_security
        - returns_policy
        - payment_options
        - comparison
        - timing_deferred
        - other
      description: >
        Structured reason for cancellation. Enum is extensible; servers
        SHOULD accept unrecognized values and treat them as "other".
    trace_summary:
      type: string
      maxLength: 500
      description: Human-readable summary of the objection.
    metadata:
      type: object
      additionalProperties:
        oneOf:
          - type: string
          - type: number
          - type: boolean
      description: >
        Flat key-value object for additional context. Arrays and nested
        objects are NOT permitted. Monetary values SHOULD be integers
        in minor units per ACS Section 3.1.
  required: [reason_code]
```

### Semantics (normative)

- The `intent_trace` is **write-only**. It MUST NOT be returned in `GET /checkout_sessions/{id}` responses.
- Cancellation MUST succeed regardless of whether the server supports or successfully processes the `intent_trace`.
- Servers SHOULD accept unrecognized `reason_code` values and treat them as `other` (forward compatibility).
- Servers SHOULD return `400 Bad Request` only for structurally malformed data (wrong types, nested objects in `metadata`). Unknown `reason_code` values are NOT errors.
- Requests with `intent_trace` MUST be idempotent. Repeated calls with the same `Idempotency-Key` MUST NOT duplicate the trace record.

### Privacy Considerations

Unlike cookie-based retargeting, Intent Traces are:

- **Explicit:** Transmitted only when the user chooses to cancel.
- **Scoped:** Sent only to the merchant involved in the session, not broadcast to ad networks.
- **Minimal:** Structured codes over free-text to reduce PII leakage.

Agents SHOULD NOT transmit PII in the `trace_summary` field unless authorized by the user.

## Files Changed

| File | Description |
|------|-------------|
| `rfcs/rfc.intent_traces.md` | Full specification with design rationale, examples, and conformance checklist |

## Notes

- Enum values are grounded in [Baymard Institute cart abandonment research](https://baymard.com/lists/cart-abandonment-rate).
- A subsequent RFC MAY define **Reconsideration Offers** keyed to specific reason codes, enabling merchants to respond to objections programmatically.

### DIFF
--- a/examples/examples.agentic_checkout.json
+++ b/examples/examples.agentic_checkout.json
@@ -406,6 +406,26 @@
       }
     ]
   },
+  "cancel_checkout_session_request": {
+    "intent_trace": {
+      "reason_code": "shipping_cost",
+      "trace_summary": "User loves the item but isn't willing to pay $10 for standard shipping.",
+      "metadata": {
+        "target_shipping_cost": 0,
+        "competitor_reference": "amazon_prime"
+      }
+    }
+  },
+  "cancel_checkout_session_request_timing_deferred": {
+    "intent_trace": {
+      "reason_code": "timing_deferred",
+      "trace_summary": "User confirmed product fit and price but plans to purchase after next paycheck.",
+      "metadata": {
+        "anticipated_timeframe": "2_weeks",
+        "deferral_trigger": "budget_cycle"
+      }
+    }
+  },
   "cancel_checkout_session_response": {
     "id": "checkout_session_123",
     "status": "canceled",
--- a/rfcs/rfc.intent_traces.md
+++ b/rfcs/rfc.intent_traces.md
@@ -1,0 +1,237 @@
+# RFC: Agentic Checkout — Intent Traces
+
+**Status:** Proposal
+**Version:** 2025-12-16
+**Scope:** Extension of `POST /cancel` to support structured intent context.
+
+This SEP extends the **Agentic Checkout Specification (ACS)** to support **Intent Tracing**. It defines the request body schema for the cancellation endpoint, enabling agents to transmit structured data regarding _why_ a checkout session was abandoned. This specification lays the groundwork for merchants to transition from probabilistic retargeting (guessing) to deterministic negotiation (bidding).
+
+---
+
+## 1. Motivation
+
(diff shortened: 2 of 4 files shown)


prasad-stripe (approved):
Looks good. Thank you for your contributions

PR #57, opened 2025-12-18
[SEP] Affiliate Attribution: Deterministic attribution transport for checkout sessions

0xtotaylor:
## Abstract

This RFC extends the `POST /checkout_sessions`, `POST /checkout_sessions/{id}`, and `POST /checkout_sessions/{id}/complete` endpoints to accept an optional `affiliate_attribution` object, enabling agents to transmit fraud-resistant attribution claims for crediting third-party publishers (affiliates) without relying on cookies, redirects, or client-side tracking.

## Motivation

Affiliate marketing is a core distribution channel for e-commerce. In traditional web flows, attribution is inferred via links, cookies, redirects, and pixels. In **agentic commerce**, the agent often performs discovery and comparison and may complete the purchase without a conventional "click trail," breaking classic attribution primitives.

### Why structured attribution?

Without a standardized attribution signal:

- Merchants cannot reliably credit publishers who influenced the purchase.
- Affiliates are disincentivized to create high-quality commerce content.
- The ecosystem devolves into opaque, platform-controlled deals rather than open, interoperable referral economics.

A structured `affiliate_attribution` object enables deterministic, fraud-resistant attribution transport from agent to merchant at checkout time.

### Why token-first?

A **provider-issued, opaque token** is the simplest way to support fraud resistance without baking any single network's API into ACP. Merchants can validate tokens out-of-band with the provider they trust, similar to how payment tokens are validated via PSPs.

### Why write-only?

Attribution often contains commercially sensitive data (publisher IDs, campaign identifiers, token secrets). The safest default is that it is **write-only**: accepted and stored by the merchant but not echoed back in responses or exposed in any read endpoint.

### Endpoints Updated

The following endpoints now accept an optional `affiliate_attribution` object in their request bodies:

- `POST /checkout_sessions` — Create Session
- `POST /checkout_sessions/{id}` — Update Session
- `POST /checkout_sessions/{id}/complete` — Complete Session

### New Schema: AffiliateAttribution

```yaml
AffiliateAttribution:
  type: object
  properties:
    provider:
      type: string
      description: Attribution provider namespace (e.g., 'impact.com').
    token:
      type: string
      description: Opaque provider-issued token for fraud-resistant validation.
    publisher_id:
      type: string
      description: Provider-scoped affiliate identifier. Required if token is omitted.
    campaign_id:
      type: string
    creative_id:
      type: string
    sub_id:
      type: string
    source:
      $ref: "#/components/schemas/AffiliateAttributionSource"
    issued_at:
      type: string
      format: date-time
    expires_at:
      type: string
      format: date-time
    metadata:
      $ref: "#/components/schemas/AffiliateAttributionMetadata"
  required: [provider]
  anyOf:
    - required: [token]
    - required: [publisher_id]
```

### Semantics (normative)

- The `affiliate_attribution` is **write-only**. It MUST NOT be returned in `GET /checkout_sessions/{id}` responses, list endpoints, or webhook payloads.
- Checkout MUST succeed regardless of whether the server supports or successfully processes the attribution.
- **Write-once with idempotency:**
  - Same `Idempotency-Key` + identical payload → success (idempotent replay)
  - Same `Idempotency-Key` + different payload → **409 Conflict**
  - Different/missing key + conflicting `provider`/`publisher_id` → first claim wins, subsequent claims silently ignored
- Servers SHOULD accept unknown `provider` values (extensible design).
- Servers SHOULD return `400 Bad Request` with code `pii_not_allowed` for payloads containing obvious PII.
- **Forward compatibility:** Servers SHOULD ignore unknown fields to support future extensions (per RFC §8.2).

### Privacy Considerations

- Attribution MUST NOT contain user PII (emails, phone numbers, addresses, stable user identifiers, device fingerprints).
- Tokens SHOULD be treated as secrets and not logged in plaintext.
- Agents MUST NOT treat `affiliate_attribution` as a ranking boost mechanism—it is strictly for post-selection attribution/settlement.

## Files Changed

| File | Description |
|------|-------------|
| `rfcs/rfc.affiliate_attribution.md` | Full specification with design rationale, examples, error handling, security considerations, and conformance checklist |
| `spec/json-schema/schema.agentic_checkout.json` | Added `AffiliateAttribution`, `AffiliateAttributionSource`, `AffiliateAttributionMetadata` schemas; extended request schemas |
| `spec/openapi/openapi.agentic_checkout.yaml` | Added schemas, updated endpoint descriptions, added examples |
| `examples/examples.agentic_checkout.json` | Added `create_checkout_session_request_with_affiliate_attribution` and `complete_checkout_session_request_with_affiliate_attribution` examples |
| `changelog/unreleased.md` | Documented the Affiliate Attribution extension |

## Notes

- This RFC is designed to be **protocol-neutral**, supporting multiple affiliate networks/providers without privileging any one business model.
- Commission rates, payout schedules, and settlement rails are explicitly out of scope—token verification happens out-of-band between merchant and provider.
- A subsequent RFC MAY expand multi-touch attribution support; this initial proposal focuses on minimal, single-touch attribution transport.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,86 @@
+# Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in the Agentic Commerce Protocol community a harassment-free experience for everyone, regardless of age, body size, visible or invisible disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, caste, color, religion, or sexual identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes, and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a professional setting
+
+## Enforcement Responsibilities
+
+Project maintainers are responsible for clarifying and enforcing our standards of acceptable behavior and will take appropriate and fair corrective action in response to any behavior that they deem inappropriate, threatening, offensive, or harmful.
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, and will communicate reasons for moderation decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when an individual is officially representing the community in public spaces. Examples of representing our community include using an official e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported to the project maintainers responsible for enforcement. See [MAINTAINERS.md](MAINTAINERS.md) for contact information.
+
+All complaints will be reviewed and investigated promptly and fairly.
+
+All project maintainers are obligated to respect the privacy and security of the reporter of any incident.
+
+## Enforcement Guidelines
+
+Project maintainers will follow these Community Impact Guidelines in determining the consequences for any action they deem in violation of this Code of Conduct:
+
+### 1. Correction
+
+**Community Impact**: Use of inappropriate language or other behavior deemed unprofessional or unwelcome in the community.
+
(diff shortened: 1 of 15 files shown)


0xtotaylor:
After discussing attribution models with various stakeholders at [Affiliate Summit West](https://www.affiliatesummit.com/west), I'd like to propose supporting attribution at **both** create and complete rather than choosing one.

### What I heard from the market

* **Networks are split:** Some use last-touch (complete), others are moving toward multi-touch/MMM models.
* **Publishers want first-touch protection:** "Compensation model is the missing link."
* **No single attribution model fits all:** Networks want flexibility to apply their own logic.

### Proposal

Allow `affiliate_attribution` on both create and complete with a `touchpoint` field:

```yaml
affiliate_attribution:
  token: string
  touchpoint: "first" | "last"
  # ... other fields

```

### Why this works

| Use case | How it's handled |
| --- | --- |
| **Last-touch** (industry default) | Use `complete` only, ignore `create` |
| **First-touch protection** | Use `create` only |
| **Multi-touch/MMM** | Capture both, network applies weighting |

### Benefits

1. **Addresses OAI's hijacking concern** — first touch is recorded at create.
2. **Preserves clean settlement** — complete still pairs with conversion.
3. **Doesn't prescribe attribution model** — protocol captures touchpoints, networks decide weighting.
4. **Future-proofs the spec** — as multi-touch becomes standard for agentic commerce, the protocol already supports it.

Happy to update the PR if this direction makes sense. Curious what @prasad-stripe, @aravindrao-openai, and others think.

aravindrao-openai:
Seems reasonable to me - LGTM!

Issue #55, opened 2025-12-17
[Feature Request & Spec Proposal] Add 3DS Authentication Flow Support


yehonatanz:
## Summary

The current ACP specification includes a `requires_3ds` error code in `MessageError`, but does not define the complete flow for handling 3DS authentications. This proposal suggests a standardized 3DS 2.x authentication flow to enable secure card payments that require issuer authentication.

## Motivation

At Forter, we facilitate safe online and agentic commerce for some of the largest merchant brands in the US and globally.
As 3DS is increasingly mandated for card-not-present transactions, supporting it becomes crucial for international commerce:
- **PSD2/SCA** (Europe) requires strong customer authentication for most online payments
- **RBI** (India) requires additional factor authentication for all online payments
- **Oct 2025 PSA** (Japan) mandates 3DS and SCA similarly to PSD2

Operating in these markets necessitates 3DS support, and I'd love to see ACP properly specify it.

### High-Level Flow

3DS has 4 phases:
1. [Regardless of transaction] Initiating the process and fingerprinting the device
2. Sending authentication request to the issuing bank (usually through an intermediary)
3. [Optional, if challenge was requested] Presenting the challenge
4. Completing the process and authorizing the payment with the 3DS results

My suggestion is as follows:
```
1. Agent calls POST /checkout_sessions/{id}/complete with payment token
2. Merchant initiates 3DS with their PSP
3. If challenge required:
   - Merchant returns status=in_progress with `requires_3ds` error + `authentication_data` (`acsURL`, `creq`, `threeDSServerTransID`, `version`)
   - Agent platform presents 3DS challenge in iframe
   - User completes authentication (OTP, biometric, etc.)
4. Agent calls POST /checkout_sessions/{id}/authenticate with `cres`
5. Merchant validates 3DS and returns `three_ds_result` (`eci`, `cavv`, etc.)
6. Agent calls POST /checkout_sessions/{id}/complete with `three_ds_proof`
7. Merchant authorizes payment and creates order
8. Returns status=completed with order
```

### Why Two-Step Completion?

The `/authenticate` endpoint validates the 3DS challenge and returns authentication proof (ECI, CAVV, DS Transaction ID), but does **not** create the order. The agent must call `/complete` again with `three_ds_proof` to finalize. This separation provides:

- **Semantic clarity**: `/complete` always creates the order, `/authenticate` only handles authentication
[...]

PR #53, opened 2025-12-12
[SEP] Adding support for consumer authentication (3DS)

elorakelkel-stripe:
## Abstract

This SEP proposes extending the Agentic Checkout APIs to enable agents and sellers to exchange the information necessary for the agent and their PSP to operate consumer authentication (specifically 3D Secure) on behalf of the seller.

## Motivation

Current Delegate Payment and Agentic Checkout API schemas do not support any form of consumer authentication.

### Why add such support on the Agentic Checkout API? 

The main motivation for supporting consumer authentication, and specifically 3D Secure, is compliance in regulated markets (for example, Strong Customer Authentication in Europe).

image

For simplicity, the fingerprinting step is omitted from the diagram, this step is optional. 

In this flow, the agent may rely on any PSP to complete 3DS authentication. 

### Why add such support on the Delegate Payment Token API? 

With agentic tokens, consumer authentication (e.g., 3D Secure) may be required when creating the delegate payment token. No changes to the Delegate Payment Token API are included in this SEP. Any proposals affecting the delegate payment token flow will be submitted in a separate PR to keep this change focused and reviewable.

### New response fields (seller → agent)

Extend CheckoutSessionBase (and CheckoutSessionWithOrder by extension) with relevant objects. authentication requirement is encoded via the session `status` enum (see below). `authentication_metadata` carries the seller-provided data needed to perform the authentication.

(1) Update `CheckoutSessionBase` `status` enum to include:

```
status:
  type: string
  enum:
    [
        not_ready_for_payment,
        authentication_required,
        ready_for_payment,
        completed,
        canceled,
        in_progress,
    ]
```

(Keep the rest of `CheckoutSessionBase` unchanged; `authentication_required` is the canonical indicator that the session requires authentication such as 3DS.)

(2) Schema: `authentication_metadata` (seller → agent)

```
AuthenticationMetadata:
      type: object
      additionalProperties: true
      description: >
        Seller-provided authentication metadata for 3DS flows.
      required:
        - channel
        - acquirer_details
        - directory_server
      properties:
        channel:
          type: object
          additionalProperties: false
          description: Details about the channel used for this 3DS Authentication.
          required:
            - type
            - browser
          properties:
            type:
              type: string
              enum: [browser]
              description: Use "browser" to indicate the transaction originated from a browser.
            browser:
              type: object
              additionalProperties: false
              description: Browser details collected server- or client-side.
              properties:
                accept_header:
                  type: string
                  description: The HTTP Accept header from the cardholder’s browser.
[...]

PR #66, opened 2025-12-25
Support Merchant 3DS Execution

yehonatanz:
Addresses #55 and in conflict with #53 

This PR proposes a standard of 3DS execution that provides merchants with as much freedom as possible.
This essentially allows merchants to bring their own 3DS provider, be it their PSP or a 3rd party, by keeping the agent's responsibility to a minimum and adhering the lowest common denominator of 3DS - the EMVco speficication.

This should play well with many real world setups, and was planned with several 3DS setups in mind (including Adyen's execution and 3rd party execution).

This sequence diagram summarizes the proposed changes:
```mermaid
sequenceDiagram
    autonumber
    actor B as Buyer
    participant A as Agent
    participant P as Agent's PSP
    participant M as Merchant
    A->>P: POST /agentic_commerce/delegate_payment<br> (payment method + allowance + billing information + risk data)
    P->>A: 200 OK + payment delegation token
    rect rgb(0, 255, 0, 0.2)
        note over B,M: This Proposal
        A->>M: POST /checkout_sessions/:id/complete (payment delegation token + buyer?)
        M->M: Decide if 3DS might be relevant<br>execute PReq with own PSP/3rd party
        M->>A: 200 OK + updated checkout session info<br>(+ pending_action=3ds_method+method_url+three_ds_server_trans_id)
        alt pending_action=3ds_method
            A->>A: Render 3DS Method
        end
        A->>M: POST /checkout_sessions/:id/complete (payment delegation token + method_completion_indicator + buyer?)
        M->M: Decide whether to execute 3DS<br>execute AReq with own PSP/3rd party
        M->>A: 200 OK + finalized checkout session info (+ pending_action=3ds_challenge+acs_url+encoded_creq)
        alt requires_3ds
            A->>A: Render 3DS Challenge
            B->>A: Complete 3DS Challenge
            A->>M: POST /checkout_sessions/:id/complete (+cres)
        end
    end
    A->>M: POST /checkout_sessions/:id/complete (payment delegation token + buyer?)
    M->>A: 200 OK + finalized checkout session info + checkout confirmed
    A->>B: Delivery is on the way!
```

### DIFF
--- a/examples/examples.agentic_checkout.json
+++ b/examples/examples.agentic_checkout.json
@@ -333,7 +333,12 @@
         "type": "return_policy",
         "url": "https://www.testshop.com/legal/return-policy"
       }
-    ]
+    ],
+    "order": {
+      "id": "ord_abc123",
+      "checkout_session_id": "checkout_session_123",
+      "permalink_url": "https://www.testshop.com/orders/ord_abc123"
+    }
   },
   "get_checkout_session_response": {
     "id": "checkout_session_123",
@@ -540,7 +545,8 @@
       ]
     }
   },
-  "complete_checkout_session_requires_3ds_response": {
+  "complete_checkout_session_requires_3ds_method_response": {
+    "_comment": "Response when 3DS Method (device fingerprinting) is required before authentication",
     "id": "checkout_session_123",
     "buyer": {
       "first_name": "John",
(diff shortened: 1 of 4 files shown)


elorakelkel-stripe:
I do not consider #53 and #66 to be in conflict; they address different requirements. Not all sellers have the capability to operate 3DS through their own provider. 

I think we should also probably consider the long-term vision in which a payment should proceed without human intervention after delegated payment tokens creation. For example, a consumer could establish a spending allowance for an agent. To avoid further authentication after token creation, we could eventually rely on 3RI authentication or agentic tokens; in that model, consumer authentication (e.g., 3D Secure) is still required when creating the delegated payment token. This is not currently in the spec, but we are drafting a proposal for it now.

yehonatanz:
Thanks @elorakelkel-stripe! The proposals definitely address different requirements. I agree it would be great to have them both in the spec (assuming we reuse data models when appropriate to reduce bloat).

Re authentication at delegated token creation time - that's a great point! That's an important capability to support for the longer term. Similarly to my previous point, I think it's best if we offered both options here as well - fully delegate authentication to the agent's PSP or bring your own, while reusing data models as much as possible. I'd be happy to workshop this, I strongly believe allowing merchants flexibility in those integration points is key to bring in enterprise grade merchants

elorakelkel-stripe:
Hi @yehonatanz,

FYI, I have left [a comment](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/55#issuecomment-3785057446) on the related issue focusing on the 3DS Method considerations and the latency implications introduced by this PR. Happy to discuss!

Issue #105, opened 2026-02-05
[SEP][Proposal]: Native Orders Support - Rich Post-Purchase Lifecycle Tracking


nvp-stripe:
# SEP: Native Orders Support

## Preamble

- **Status**: `proposal`
- **Type**: Major Change
- **Created**: 2026-02-05

---

## Abstract

This SEP proposes enhancing the existing `Order` schema with rich post-purchase lifecycle tracking capabilities. The enhancement introduces:

- **Per-line-item quantity tracking** — `quantity.ordered` and `quantity.shipped` fields enabling visibility like "2 of 3 items have shipped"
- **Fulfillment tracking** — a `fulfillments[]` array covering shipping, in-store pickup, and digital delivery, each with an event history timeline
- **Adjustments** — a unified `adjustments[]` model for refunds, store credits, returns, disputes, and chargebacks

All new fields are optional, maintaining full backward compatibility. Merchants can progressively adopt richer fields as their systems support them. This allows agents to answer common post-purchase questions without redirecting buyers to the merchant's website.

---

## Motivation

ACP is the **Agentic Commerce Protocol** — not just checkout. Agents need to answer post-purchase questions, but the current minimal `Order` schema only contains three required fields (`id`, `checkout_session_id`, `permalink_url`) and lacks the data for meaningful responses.

### The Problem

| Buyer Question | Current Capability | Impact |
|----------------|-------------------|--------|
| "Where's my order?" | ❌ Redirect to permalink | Poor agent experience |
| "What did I order?" | ❌ No line items in Order | Can't itemize |
| "Which items shipped?" | ❌ No quantity tracking | Can't give partial updates |
| "Did my refund go through?" | ❌ No adjustment visibility | Buyer left in the dark |

### Why This Matters

Post-purchase is where buyer anxiety peaks. "Where's my package?" is one of the most common customer service inquiries. An agent that can only say "please check this link" provides minimal value compared to one that can say:

> "Your Running Shoes were delivered yesterday. Your T-Shirts are backordered and will ship February 15th. I see your $12.50 refund for the damaged item was processed on February 3rd."

This level of visibility is table stakes for modern commerce. Agents need it.

---

### New Schemas

| Schema | Purpose |
|--------|---------|
| `OrderLineItem` | Line item with title, image, price, and quantity tracking |
| `OrderLineItemQuantity` | Object with `ordered` (required) and `shipped` (optional) fields |
[...]

PR #106, opened 2026-02-05
SEP: Add rich post-purchase lifecycle tracking to Order schema

nvp-stripe:
# Native Orders Support

## 📋 SEP Metadata

- **SEP Number**: #105
- **Status**: `proposal`
- **Related Issues/RFCs**: `rfcs/rfc.orders.md`

---

## 🎯 Abstract

This SEP enhances the existing `Order` schema with rich post-purchase lifecycle tracking. It introduces:

- **Per-line-item quantity tracking** (`ordered`/`shipped`) — enabling "2 of 3 items have shipped" visibility
- **Fulfillment tracking** — covering shipping, pickup, and digital delivery with event history
- **Adjustments** — a unified model for refunds, credits, returns, disputes, and chargebacks

All new fields are optional, maintaining full backward compatibility while enabling merchants to provide richer order experiences. This allows agents to answer post-purchase questions like "Where's my order?", "Which items shipped?", and "Did I get a refund?" without requiring buyers to visit the merchant's website.

---

## 💡 Motivation

Agents need to answer post-purchase questions, but the current minimal `Order` schema (id, checkout_session_id, permalink_url) lacks the data to do so:

| Question | Current State | With This SEP |
|----------|---------------|---------------|
| "Where's my order?" | ❌ Must redirect to permalink | ✓ `fulfillments[]` with tracking |
| "What did I order?" | ❌ No line items in Order | ✓ `line_items[]` with details |
| "Which items shipped?" | ❌ No tracking | ✓ `quantity.shipped` per item |
| "I got a refund?" | ❌ No visibility | ✓ `adjustments[]` with status |

ACP is the **Agentic Commerce Protocol**, not just checkout. Post-purchase order tracking is a core commerce capability that agents need.

---

### New Schemas

| Schema | Description |
|--------|-------------|
| `OrderLineItem` | Per-item tracking with quantity breakdown |
| `OrderLineItemQuantity` | Quantity tracking with `ordered` and `shipped` fields |
| `LineItemReference` | References line items in fulfillments and adjustments |
| `Fulfillment` | Shipping, pickup, or digital delivery tracking |
| `FulfillmentEvent` | Append-only delivery event log |
| `Adjustment` | Refunds, credits, returns, disputes, chargebacks |
| `OrderTotals` | Order-level financial summary |

### Enhanced Order Schema

```yaml
Order:
  properties:
    # Existing required
    id: { type: string }
    checkout_session_id: { type: string }
    permalink_url: { type: string, format: uri }
    
    # Existing optional (unchanged)
    order_number, status, estimated_delivery, confirmation, support
    
    # NEW optional
    line_items: { type: array, items: OrderLineItem }
    fulfillments: { type: array, items: Fulfillment }
    adjustments: { type: array, items: Adjustment }
    totals: { $ref: OrderTotals }
```

### Files Updated

- `spec/unreleased/openapi/openapi.agentic_checkout.yaml`
- `spec/unreleased/json-schema/schema.agentic_checkout.json`
- `rfcs/rfc.orders.md` (new)
- `examples/orders/` (new directory with 4 examples)
- `changelog/unreleased/native-orders-support.md`

---

### Design Decisions

| Decision | Rationale |
|----------|-----------|
| **`fulfillments[]` with type discriminator** | Covers all delivery methods: shipping, in-store pickup, and digital download. A single unified model means agents don't need special-case logic. |
| **`quantity.ordered` and `quantity.shipped`** | Clear, unambiguous semantics. "Shipped" means handed to carrier — merchants already think in these terms. Easily extensible to `delivered` or `returned` in the future. |
| **`adjustments[]` as unified model** | Post-order financial changes come in many forms: refunds, store credits, returns, disputes, chargebacks. A single array with type discriminator handles all of them cleanly. |
| **Fixed enums for types** | Predictable for agents. Closed enums enable reliable automation without brittle string parsing. |
| **Events nested in Fulfillment** | Each fulfillment has its own timeline (label_created → shipped → delivered). Nesting events under their fulfillment is the natural structure. |

---

## 🔄 Backward Compatibility

**No breaking changes.** All new fields are optional.

- Existing integrations that only read `id`, `checkout_session_id`, and `permalink_url` continue to work unchanged
- Merchants can progressively adopt richer fields as their systems support them
- Empty arrays (`fulfillments: []`, `adjustments: []`) are valid

---

Not yet available. This PR includes:
- Complete OpenAPI and JSON Schema definitions
- RFC documentation with detailed field descriptions
- 4 example JSON files demonstrating common scenarios

---

## 🔒 Security Implications

- No new authentication/authorization requirements
- Order data should already be protected by existing merchant authentication
- No PII additions beyond what's already in checkout responses

---

## ✅ Pre-Submission Checklist

- [x] This PR includes a changelog entry file in `changelog/unreleased/native-orders-support.md`

---

### Example: Partial Shipment

```json
{
  "id": "ord_123",
  "checkout_session_id": "cs_456",
  "permalink_url": "https://merchant.com/orders/123",
  "line_items": [
    { "id": "li_shoes", "title": "Running Shoes", "quantity": { "ordered": 3, "shipped": 3 }, "status": "shipped" },
    { "id": "li_shirts", "title": "T-Shirt", "quantity": { "ordered": 2, "shipped": 0 }, "status": "processing" }
  ],
  "fulfillments": [
    { "id": "ful_1", "type": "shipping", "status": "delivered", "line_items": [{ "id": "li_shoes", "quantity": 3 }], "tracking_number": "12345" },
    { "id": "ful_2", "type": "shipping", "status": "pending", "line_items": [{ "id": "li_shirts", "quantity": 2 }], "description": "Backordered - ships Feb 15" }
  ]
}
```

**Agent response:** "Your 3 pairs of Running Shoes were delivered. Your 2 T-Shirts are backordered and will ship Feb 15."

---

## 🙋 Questions for Reviewers

1. Should `status` be required on line items, or is derivation from `quantity.shipped` sufficient?
2. Should we add `quantity.delivered` now, or wait for future extensibility?
3. Are the FulfillmentEvent type enum values comprehensive enough?

### DIFF
--- a/rfcs/rfc.orders.md
+++ b/rfcs/rfc.orders.md
@@ -144,6 +144,7 @@ The existing `Order` schema gains optional fields:
 | `quantity` | OrderLineItemQuantity | Yes | Quantity tracking |
 | `unit_price` | integer | No | Price per unit (minor units) |
 | `subtotal` | integer | No | Line total (minor units) |
+| `totals` | Total[] | No | Optional line-item totals breakdown (same schema as checkout) |
 | `status` | enum | No | Derived line item status |
 
 **OrderLineItemQuantity:**
--- a/spec/unreleased/json-schema/schema.agentic_checkout.json
+++ b/spec/unreleased/json-schema/schema.agentic_checkout.json
@@ -2161,6 +2161,11 @@
           "type": "integer",
           "description": "Line total in minor currency units"
         },
+        "totals": {
+          "type": "array",
+          "items": { "$ref": "#/$defs/Total" },
+          "description": "Optional line-item level totals breakdown using the same Total schema as checkout. Merchants who can provide richer breakdowns MAY use this alongside or instead of unit_price/subtotal."
+        },
         "status": {
           "type": "string",
           "enum": ["processing", "partial", "shipped", "delivered", "canceled"],
--- a/spec/unreleased/openapi/openapi.agentic_checkout.yaml
+++ b/spec/unreleased/openapi/openapi.agentic_checkout.yaml
@@ -1312,6 +1312,10 @@ components:
         subtotal:
           type: integer
           description: "Line total in minor currency units (quantity.ordered * unit_price)"
+        totals:
+          type: array
+          items: { $ref: "#/components/schemas/Total" }
+          description: "Optional line-item level totals breakdown using the same Total schema as checkout. Merchants who can provide richer breakdowns MAY use this alongside or instead of unit_price/subtotal."
         status:
           type: string
           enum: [processing, partial, shipped, delivered, canceled]


kxbnb (commented):
Went through the schemas, RFC, and examples. The overall design is solid — progressive enrichment with all-optional fields is the right approach. A few things to sort out before this can land:

### Order.status enum doesn't match webhook EventDataOrder.status

The PR adds `[confirmed, processing, shipped, delivered, canceled]` to Order.status. But the existing webhook spec (`openapi.agentic_checkout_webhook.yaml`) uses a completely different enum on `EventDataOrder.status`: `[created, manual_review, confirmed, canceled, shipped, fulfilled]`.

- `created` and `manual_review` are in the webhook but not in Order
- `delivered` and `processing` are in Order but not in the webhook
- `fulfilled` is in the webhook but not in Order (is this `delivered`?)

These should converge, or at least the RFC should document the mapping.

### EventDataOrder is a separate schema from Order

Related: the RFC says "merchants MUST include the full order object" in webhook payloads. But the webhook spec doesn't use the `Order` schema at all. `EventDataOrder` is its own schema with only `type, checkout_session_id, permalink_url, status, refunds`. None of the new fields (line_items, fulfillments, adjustments, totals) will show up in webhooks unless `EventDataOrder` is updated or replaced with a `$ref` to `Order`.

This is the bigger version of the `refunds[]` vs `adjustments[]` overlap I flagged on the issue. The webhook schema needs to be part of this PR, otherwise the "full order object" statement in the RFC is aspirational rather than enforced.

### Fulfillment statuses don't all apply to non-shipping types

`Fulfillment.status` has `[pending, processing, shipped, in_transit, out_for_delivery, delivered, failed, canceled]`. But for `type: "pickup"`:
- `shipped`, `in_transit`, `out_for_delivery` don't apply
- Something like `ready_for_pickup` is missing

And for `type: "digital"`:
- `shipped`, `in_transit`, `out_for_delivery` don't apply
- The useful states are more like `pending`, `processing`, `delivered` (or `available`)

Two options: type-specific status enums (complex) or document which statuses apply to which types and accept that some combinations are nonsensical (simpler, my preference).

### Adjustment.amount: pre-tax or inclusive?

In the refund example, `unit_price: 29900` and adjustment `amount: 29900`. If tax was charged on that item, is the refund amount before or after tax? An agent telling the buyer "you've been refunded $299" when the actual refund includes tax ($299 + tax) is misleading. The `amount` field description should specify.

### Examples directory

Existing unreleased examples live in `examples/unreleased/`. This PR creates `examples/orders/`. Should these follow the existing convention? Or is this intentionally a new pattern for non-checkout examples?

### Minor: existing webhook example has a type bug

Not from this PR, but while reading the webhook spec: the inline example shows `amount: "1.00"` (string) while the `Refund` schema defines `amount` as `type: integer`. Might be worth fixing while you're touching this area.

nvp-stripe:
Thanks for the thorough review - really appreciate you going through both the schemas and the webhook spec side by side. Pushed updates that should address everything, going through each point:

**Order.status enum divergence**

Good catch. Extended the Order enum to `[created, confirmed, manual_review, processing, shipped, delivered, canceled]` which is now a superset of what the webhook spec had. `fulfilled` from the old webhook spec maps to `delivered` which I think is clearer semantically. Since EventDataOrder now refs Order directly (see below), theres only one source of truth for this going forward.

**EventDataOrder as separate schema**

This was the big one and you're right that it needed to be part of this PR. `EventDataOrder` now uses `allOf` with a `$ref` to the Order schema from the checkout spec, so all the new fields (line_items, fulfillments, adjustments, totals) flow through to webhooks automatically. Also removed the `Refund` schema and `refunds[]` field entirely rather than deprecating - `adjustments[]` with `type: "refund"` replaces it. Didn't see a reason to carry dead weight for backward compat when we can make a clean break here.

Updated all the inline webhook examples to show full Order payloads with the new fields.

**Fulfillment statuses for non-shipping types**

Went with your simpler suggestion - single enum with documentation on which statuses apply to which type. Added `ready_for_pickup` for the pickup case. The RFC now has a full applicability table (shipping/pickup/digital columns) and the OpenAPI description on the status field lists the valid combos per type.

**Adjustment.amount tax semantics**

Good call, this was ambiguous. `amount` is now documented as "total amount credited to the buyer, inclusive of any applicable tax" across the schema and the RFC. The refund examples show this explicitly too, eg the description says "(includes $26.16 tax)".

**totals.total ambiguity** (from the SEP comment)

Added `amount_refunded` to OrderTotals and pinned `total` as always the original charged amount at checkout. The RFC has a note explaining agents should surface both numbers rather then subtracting. Think this is the cleanest approach - agents can say "your order was $670.99 and you've been refunded $325.16" without doing arithmetic.

**Digital fulfillment fields** (from the SEP comment)

Added a `digital_delivery` sub-object on Fulfillment with `access_url`, `license_key`, and `expires_at`. Also added a new `digital-fulfillment.json` example showing a software license delivery end to end.

**Examples directory**

Didn't move these to `examples/unreleased/` - I think `examples/orders/` makes more sense as a pattern for non-checkout examples since these aren't tied to a specific release cycle the same way. 

**Webhook amount type bug**

Fixed as part of rewriting the examples - all amounts are now integers in minor units.

**Q1 - line item status**

Agreed with keeping it optional. `required` on OrderLineItem is just `[id, title, quantity]` and the RFC documents how to derive status from `quantity.shipped`.

Let me know if anything looks off or if you want to dig into any of these further.

nvp-stripe:
@prasad-stripe @aravindrao-openai - could you please take a look?

kxbnb:
@nvp-stripe - fixes look good. `allOf` for EventDataOrder and dropping `Refund` entirely are both clean calls.

+1 on @aravindrao-openai's point about reusing `Total[]`. The flat `OrderTotals` object is a second shape for the same concept. If Order totals use `Total[]`, agents parsing checkout and order responses can use the same code path.

Same thing at the line item level. Checkout spec has `totals: Total[]` on line items, but `OrderLineItem` uses flat `unit_price`/`subtotal` fields. Worth aligning those too.

nvp-stripe:
@aravindrao-openai - good catch on the totals consistency, and thanks for flagging it. I think you're right that having two shapes for the same concept is going to be annoying for agent developers - agreed we should fix this.

**Order-level totals: switching to `Total[]`:**

I'll drop the flat `OrderTotals` object and reuse `Total[]` on `Order.totals`, same as checkout sessions and fulfillment options. For `amount_refunded` I'll add it as a new value in the `Total.type` enum. It's slightly different from the other entries conceptually (post-purchase reduction vs charge composition), but the `Total` schema is really just "labeled amount with a type discriminator" and this fits fine. The `display_text` and description fields give enough room to clarify semantics.
So an order's totals would look like:

```json
"totals": [
  { "type": "subtotal", "display_text": "Subtotal", "amount": 9900 },
  { "type": "fulfillment", "display_text": "Shipping", "amount": 500 },
  { "type": "tax", "display_text": "Tax", "amount": 860 },
  { "type": "total", "display_text": "Total", "amount": 11260 },
  { "type": "amount_refunded", "display_text": "Refunded", "amount": 10760 }
]
```

This also means merchants who already produce Total[] for checkout can reuse the same serialization for order webhooks, which is a nice side effect.


@kxbnb - thanks for the helpful review! I'd push back a bit on aligning `OrderLineItem` with checkout's `LineItem.totals: Total[]` though. These serve pretty different purposes - checkout LineItem is a rich cart item with catalog data, discount breakdowns, tax details. `OrderLineItem` is a post-purchase tracking record where the interesting fields are `quantity.ordered`, `quantity.shipped`, and `status`. The `unit_price` and `subtotal` are there as lightweight reference amounts so an agent can say "your $99 shoes shipped" - not to reconstruct a full per-item tax breakdown.

I think one practical concern is merchant burden. Most major commerce platforms represent order line item pricing as flat fields (unit_price, total, tax) rather than typed totals arrays. Asking merchants to produce Total[] per line item means they'd need to reshape data they don't natively store in that format. If you feel strongly about this I could add an optional `totals: Total[]` to OrderLineItem alongside the existing fields as a forward-looking thing, but I wouldn't want to make it required or drop unit_price/subtotal.

Let me know what you think and I'll push the totals update.

prasad-stripe:
I liked the idea of adding an optional totals: Total[] to OrderLineItem alongside the existing fields as a forward-looking thing.

prasad-stripe (approved):
Thank you @nvp-stripe for detailed SEP, Thank you @aravindrao-openai and @kxbnb for thoughtful suggestions.

Issue #120, opened 2026-02-08
[SEP][Proposal]: Idempotency Requirements and Guarantees


dean-harel-nekuda-ai:
## Preamble

- **Author(s)**: Dean Harel (@dean-harel-nekuda-ai)
- **Status**: `proposal` (awaiting sponsor)
- **Type**: Major Change
- **Related**: [#97 — Specify idempotency key TTL guidance and response headers](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/97) (superseded by this SEP)

---

## Abstract

This SEP makes `Idempotency-Key` mandatory for all mutation endpoints, defines three error codes with IETF-aligned status codes (400/422/409), and consolidates key format, TTL, request equivalence, and replay behavior into a single canonical reference — across both Checkout and Delegate Payment APIs.

---

## Motivation

Agents are not humans clicking retry buttons. They are automated systems making API calls in loops, with retry logic, timeouts, and concurrent requests. For a protocol where agents autonomously execute financial transactions, "RECOMMENDED" idempotency is insufficient — a single duplicate charge due to a network retry is a trust-destroying event.

Agents need:

- **Deterministic replay safety** — a retry MUST be safe, not a gamble
- **Clear failure modes** — distinguish "you forgot the key" from "you sent conflicting data" from "original request is still processing"
- **Known retry windows** — how long is a key valid?
- **Unambiguous equivalence** — what exactly is compared when checking for duplicates?

### Current State

| Aspect                    | Checkout                                        | Delegate Payment   |
| ------------------------- | ----------------------------------------------- | ------------------ |
| `Idempotency-Key`         | RECOMMENDED                                     | RECOMMENDED        |
| Conflict detection (409)  | Described in prose; no 409 response in OpenAPI  | Yes (with example) |
| `Error.type` enum         | `request_not_idempotent` (undefined, ambiguous) | Not present        |
| `Error.code`              | Open string (implementation-defined)            | Closed enum        |
| TTL guidance              | None                                            | None               |
| Key format                | None                                            | None               |
| "Same request" definition | None                                            | None               |

`request_not_idempotent` — modeled on Stripe's `idempotency_error` — introduced ambiguity: "this request is not idempotent" (stop retrying) vs. "this request lacks idempotency protection" (retry with a key) are opposite recovery actions. No other API in the industry uses this term.

---

### `Idempotency-Key` Required

[...]

Issue #97, opened 2026-02-03
[Proposal] Specify idempotency key TTL guidance and response headers


dean-harel-nekuda-ai:
## Summary

ACP specifies idempotency semantics (RFC Section 6) but is silent on how long servers should store idempotency key mappings. Without TTL guidance, agents cannot reliably know when it's safe to retry with a new key versus expect idempotent behavior.

## Motivation

RFC Section 6 documents idempotency behavior well:

> Clients **SHOULD** set `Idempotency-Key` on create and complete requests
> Same key + identical parameters → **MUST** return original result
> Same key + different parameters → **MUST** return 409 Conflict

However, the spec is silent on:

1. How long servers should store idempotency key mappings
2. Whether TTL should be communicated to clients
3. Behavior when a key is reused after TTL expiration

### Industry Context

Different services use different TTLs:
- Stripe: 24 hours
- PayPal: 72 hours
- AWS: Varies by service

Without guidance, ACP implementations will have inconsistent retry windows, making it difficult for agents to implement reliable retry logic.

### Precedent in ACP

The Delegate Payment RFC ties token expiry to `allowance.expires_at`. Checkout sessions have `expires_at`. Idempotency keys should follow a similar pattern—their validity is naturally bounded by the session they relate to.

## Proposal

Add the following to RFC Section 6 (Idempotency):

> Servers **MUST** store idempotency key mappings for at least the **longer of**:
> - The associated session's `expires_at` plus 24 hours, or
> - 24 hours from first use of the key
>
> Servers **SHOULD** include response headers indicating idempotency key validity:
> - `Idempotency-Key-Expires`: RFC 3339 timestamp when the key mapping expires
> - `Idempotency-Key-TTL`: Seconds remaining until expiration (alternative to absolute timestamp)
>
> After TTL expiration, the key **MAY** be treated as new (no conflict detection guaranteed).

### Example Response Headers

```http
HTTP/1.1 201 Created
Content-Type: application/json
Idempotency-Key: req_abc123
Idempotency-Key-Expires: 2026-02-04T12:00:00Z
Idempotency-Key-TTL: 86400
```

### Schema Addition

Add to response headers documentation:

```yaml
Idempotency-Key-Expires:
  type: string
  format: date-time
  description: >
    RFC 3339 timestamp when the idempotency key mapping expires.
    After this time, the key may be treated as new.

Idempotency-Key-TTL:
  type: integer
  minimum: 0
  description: >
[...]

PR #121, opened 2026-02-08
SEP: Mandatory Idempotency Requirements and Guarantees

dean-harel-nekuda-ai:
## 📋 SEP Metadata

- **SEP Number**: #120
- **Author(s)**: Dean Harel (@dean-harel-nekuda-ai)
- **Related Issues/RFCs**: #120, #97 (superseded by this SEP)
- **Closes**: #120

---

## 🎯 Abstract

Agents retrying failed or timed-out POST requests currently have no protocol-level guarantee of safe replay. The `Idempotency-Key` header is optional, and there are no standardized error responses for idempotency violations. This SEP makes `Idempotency-Key` mandatory on all POST requests (max 255 characters, UUID v4 recommended, scoped to authenticated identity + endpoint) and defines three explicit error codes: `idempotency_key_required` (400), `idempotency_in_flight` (409 with `Retry-After`), and `idempotency_conflict` (422). A new `Idempotent-Replayed` response header signals cached replays. The `request_not_idempotent` error type is removed in favor of specific codes under `invalid_request`.

---

## 💡 Motivation

- **At-least-once delivery**: Network failures between agent and merchant are inevitable. Without mandatory idempotency, retried `complete` calls risk double-charging buyers.
- **No standard error contract**: Different merchants handle duplicate keys inconsistently. Agents cannot distinguish "still processing" from "payload mismatch" without standardized codes.
- **Replay transparency**: Agents need a signal (`Idempotent-Replayed`) to know whether a 2xx response is fresh or cached, enabling correct logging and UI behavior.

---

## 📐 Specification

This PR includes:

### OpenAPI Changes

- `IdempotencyKey` parameter: `required: true`, `maxLength: 255`, updated description and UUID example
- Moved `IdempotencyKey` from path-level to POST-only on `/checkout_sessions/{id}` (prevents GET inheritance)
- `Idempotent-Replayed` response header on all POST 2xx responses
- Explicit `400`, `409`, `422` responses on all POST endpoints with idempotency error examples
- Removed `request_not_idempotent` from `Error.type` enum
- Updated `409` on delegate payment from payload mismatch to in-flight collision with `Retry-After`
- Added `idempotency_key_required` and `idempotency_in_flight` to delegate payment `Error.code` enum (OpenAPI + JSON schema)

### JSON Schema Changes

- Removed `request_not_idempotent` from checkout `Error.type` enum

### Examples

- Added 3 idempotency error examples to checkout examples
- Added 2 idempotency error examples to delegate payment examples
- Updated existing `idempotency_conflict` message for clarity

### RFC Prose Changes

- **`rfc.agentic_checkout.md`**: Rewrote §6 (Idempotency, Retries & Concurrency) from 14-line summary to full normative specification (§6.1–6.8) covering key requirement, request equivalence semantics, replay behavior with `Idempotent-Replayed` header, error responses (400/422/409), 5xx caching prohibition, 24-hour key retention, extension field participation, and implementation guidance
- **`rfc.agentic_checkout.md`**: Fixed cross-references — `Idempotency-Key` RECOMMENDED → REQUIRED in header listing, response header echo updated, removed `request_not_idempotent` from `Error.type` enum, updated conformance checklist to reference §6
- **`rfc.delegate_payment.md`**: Rewrote §5 (Idempotency & Retries) from 13-line summary to full normative specification (§5.1–5.7) mirroring checkout structure minus extension fields and update-endpoint semantics
- **`rfc.delegate_payment.md`**: Fixed cross-references — `Idempotency-Key` SHOULD → MUST in preparation phase, RECOMMENDED → REQUIRED in header listing, added `idempotency_key_required` and `idempotency_in_flight` to `Error.code` enum, updated conformance checklist to reference §5

### Changelog

- `changelog/unreleased/idempotency-requirements.md`

---

## 🤔 Rationale

- **Required vs optional**: Making idempotency optional creates a two-tier ecosystem where some merchants support safe retries and others don't. Mandatory keys eliminate this ambiguity.
- **409 for in-flight, 422 for mismatch**: The current spec returns 409 for both payload mismatch and in-flight collision — errors with opposite recovery actions (stop vs. retry). The 422/409 split puts the retryability signal at the HTTP status code layer, so agents can make retry decisions without parsing the error body.
- **Removing `request_not_idempotent`**: A dedicated error type was unnecessary; specific codes under `invalid_request` are more granular and consistent with Stripe/Shopify conventions.
- **255 char max**: Generous enough for UUIDs and composite keys while preventing abuse.

---

## 🔄 Backward Compatibility

Three breaking changes:

1. **`Idempotency-Key` RECOMMENDED → MUST** — Clients not sending the header will receive `400` with `code: idempotency_key_required`, providing a clear upgrade signal.
2. **`request_not_idempotent` removed from `Error.type` enum** — Replaced by `type: invalid_request` + specific `code` values. Any client switching on this type value needs to update.
3. **409 split into 422 (payload mismatch) and 409 (in-flight only)** — Clients handling 409 for payload mismatch must now handle 422. The status code itself carries the retryability signal (422 = permanent, 409 = transient).

---

This PR is the reference implementation. It includes OpenAPI spec changes, JSON schema updates, example error responses, normative RFC prose (§6 checkout, §5 delegate payment), and a changelog entry — for both Agentic Checkout and Delegate Payment APIs.

---

## 🔒 Security Implications

- **Replay protection**: Mandatory idempotency keys eliminate the class of vulnerabilities where duplicate charges occur due to missing keys.
- **Key predictability**: Sequential or predictable keys could allow an attacker to probe for existing transactions. UUID v4 (128 bits of randomness) is recommended to prevent this.
- **Key scoping**: Keys MUST be scoped to authenticated identity + endpoint to prevent cross-client collision attacks.
- **Retry-After**: The 409 response with `Retry-After` prevents thundering-herd retries.

---

## ✅ Pre-Submission Checklist

- [x] I have discussed this proposal in the community — via issue #97, with interest from @prasad-stripe
- [x] This PR includes a changelog entry file in `changelog/unreleased/`

---

## 📚 Additional Context

Idempotency patterns follow established conventions from Stripe, Shopify, and IETF [`draft-ietf-httpapi-idempotency-key-header-07`](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/).

### Design Constraints

- **`Error.type` is locally enforced per-schema, not protocol-wide.** Checkout and Delegate Payment each define their own closed `Error.type` enum, and these enums diverge (`request_not_idempotent` in Checkout only, `rate_limit_exceeded` in Delegate Payment only). This SEP's idempotency errors use `invalid_request`, which exists in both.
- **`Error.code` compounds the enforcement gap.** Checkout uses an open string (implementation-defined); Delegate Payment uses a closed enum. This SEP adds well-known idempotency codes to Delegate Payment's enum. Checkout's open string means schema validation alone does not enforce their use — compliance relies on spec prose. Broader error harmonization is out of scope for this SEP.
- **Capabilities cannot communicate idempotency parameters.** ACP's `capabilities` object is returned in responses, not available before the first request. This creates a chicken-and-egg problem: the session-creation request that would receive capabilities itself needs idempotency protection. Delegate Payment is a standalone endpoint with no session — capabilities never apply. All idempotency parameters are therefore protocol-level mandates, not negotiated capabilities. ACP negotiates business capabilities (what can we do together?) but mandates protocol safety properties (how do we interact safely?).

---

## 🙋 Questions for Reviewers

---

**Note**: SEPs require unanimous approval from Founding Maintainers. See [governance.md](../docs/governance.md) and [sep-guidelines.md](../docs/sep-guidelines.md) for the full process.

### DIFF
--- a/rfcs/rfc.agentic_checkout.md
+++ b/rfcs/rfc.agentic_checkout.md
@@ -236,7 +236,7 @@ All money fields are **integers (minor units)**.
   | Variation | Equivalent? |
   |---|---|
   | Different key ordering | Yes |
-  | `null` value vs absent key | Servers **SHOULD** treat as equivalent |
+  | `null` value vs absent key | **Different** — `null` means "clear this field"; absent means "do not modify" |
   | Trailing zeros in numbers (`1.0` vs `1`) | Yes |
   | Array element ordering | **No** — arrays are order-sensitive |
 - Monetary values **SHOULD** use **string** or **integer-cent** representations to avoid floating-point ambiguity.
@@ -296,6 +296,7 @@ All idempotency errors use `type: "invalid_request"` and the following codes:
 - **Gateway / middleware pattern:** Idempotency logic is **RECOMMENDED** to live in a middleware layer in front of business logic, so all endpoints gain consistent behavior.
 - **Atomic transaction boundaries:** The idempotency key record and the business operation **SHOULD** be committed in the **same ACID transaction** to prevent ghost keys (key stored, operation failed) or lost keys (operation succeeded, key not stored).
 - **Recovery points:** When an operation triggers foreign state mutations (e.g., PSP authorization), servers **SHOULD** implement recovery-point semantics — record intermediate state so that retries can resume rather than restart.
+- **SDK serialization:** Because `null` and absent are semantically distinct (§6.2), SDK authors **SHOULD** ensure their serializers preserve the distinction. Clients **MUST** only include `null` when they intend to clear a field, and **MUST** omit the key when they intend to leave the field unchanged.
 
 ---
 
--- a/rfcs/rfc.delegate_payment.md
+++ b/rfcs/rfc.delegate_payment.md
@@ -210,7 +210,7 @@ Exactly **one** credential type is supported today: **card**.
   | Variation | Equivalent? |
   |---|---|
   | Different key ordering | Yes |
-  | `null` value vs absent key | Servers **SHOULD** treat as equivalent |
+  | `null` value vs absent key | **Different** — `null` means "clear this field"; absent means "do not modify" |
   | Trailing zeros in numbers (`1.0` vs `1`) | Yes |
   | Array element ordering | **No** — arrays are order-sensitive |
 - Monetary values **SHOULD** use **string** or **integer-cent** representations to avoid floating-point ambiguity.
@@ -265,6 +265,7 @@ All idempotency errors use `type: "invalid_request"` and the following codes:
 - **Gateway / middleware pattern:** Idempotency logic is **RECOMMENDED** to live in a middleware layer in front of business logic.
 - **Atomic transaction boundaries:** The idempotency key record and the tokenization operation **SHOULD** be committed in the **same ACID transaction**.
 - **Recovery points:** When tokenization triggers external PSP calls, servers **SHOULD** implement recovery-point semantics — record intermediate state so that retries can resume rather than restart.
+- **SDK serialization:** Because `null` and absent are semantically distinct (§5.2), SDK authors **SHOULD** ensure their serializers preserve the distinction. Clients **MUST** only include `null` when they intend to clear a field, and **MUST** omit the key when they intend to leave the field unchanged.
 
 ---
 
@@ -371,13 +372,13 @@ All idempotency errors use `type: "invalid_request"` and the following codes:
 
 ### 8.3 Error (idempotency conflict)
 
-**409 Conflict**
+**422 Unprocessable Entity**
 
 ```json
 {
   "type": "invalid_request",
   "code": "idempotency_conflict",
-  "message": "Same Idempotency-Key used with different parameters"
+  "message": "Idempotency-Key has already been used with a different request body"
 }
 ```
 


dean-harel-nekuda-ai:
> Thanks @dean-harel-nekuda-ai for this PR — we’re largely aligned with the direction. I will be happy to sponsor this change. Making Idempotency-Key required on all POSTs and defining clear equivalence (body-only), replay behavior, and the three error cases (400/422/409) gives agents and servers a solid, predictable contract. The 5xx-no-cache and 24h retention rules are good defaults, and the implementation guidance (middleware, atomic commit, recovery points) is helpful. A few small, non-blocking nits we’d consider before or right after merge:
> 
> 1. Changelog date: The RFC changelog uses 2026-02-08; Let's keep it unreleased, we will be clubbing a bunch of non-backward compatible changes to cut a release.
> 2. null vs absent: The “implementation-defined (pick one, document it)” for null vs absent could stay as-is or be tightened to “SHOULD treat as equivalent” if we want more consistency across implementations.
> 3. Example vs RFC wording: Examples use “Idempotency-Key reused with different request body” and the RFC uses “Idempotency-Key has already been used with a different request body” — optional to align for consistency.
> 
> Otherwise we’re good with this. Thanks for driving the idempotency spec.

@prasad-stripe thanks again for the quick turnaround and feedback. Addressed all.

dean-harel-nekuda-ai:
@prasad-stripe Quick update on nit no. 2 (null vs absent): after further consideration I'm reverting this to **Different** instead of SHOULD-equivalent. ACP's update endpoints rely on `null` = "clear field" vs absent = "don't modify" (consistent with Square, Stripe v2, and RFC 7396 JSON Merge Patch). Collapsing them at the idempotency layer would contradict the protocol's own update semantics. Full rationale with vendor research is posted on #120.

Added SDK serialization guidance in §6.8 / §5.7 to make the expectation explicit. Also fixed a stale example in delegate_payment §8.3 (was 409 with old wording, now 422 per §5.4). The other nits (changelog dates, message alignment) remain as addressed.

Issue #136, opened 2026-02-12
[SEP][Proposal] SEP: Discovery Capabilities Endpoint


nvp-stripe:
- **Status**: `proposal`
- **Type**: [x] Major Change [ ] Process Change
- **Related Issues/RFCs**: [#15 (ACP Manifest and Agent Whitelisting Flow)](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/15), `rfcs/rfc.capability_negotiation.md`, `rfcs/rfc.discovery.md`

---

## Abstract

This SEP adds a `GET /capabilities` endpoint to ACP that returns platform-level information (protocol version, supported API versions, available extensions, intervention types, services, currencies, and locales) without requiring authentication. It lets agents answer "does this merchant support ACP, and what features are available?" before creating a checkout session.

This does not replace inline capability negotiation. Merchant-specific and session-specific capabilities like payment methods and payment handlers stay in the `capabilities` object on `POST /checkout_sessions`. Discovery is about stable, platform-scoped metadata that is the same for all merchants on the platform.

---

### Problems Being Solved

ACP currently has no way for an agent to check whether a merchant supports ACP without trying to create a checkout session. This creates several problems:

1. **Blind first requests**: Agents have to call `POST /checkout_sessions` and interpret failure responses (404, auth errors, malformed JSON) to figure out if a merchant even speaks ACP. This wastes API calls and creates unnecessary sessions.

2. **Version guessing**: There is no way to learn the supported API version before making a request. The first call may fail with a version mismatch, requiring a retry with a different `API-Version` header.

3. **No feature overview**: Agents cannot know which high-level features a platform offers (orders, discounts, delegate payment) without starting a transaction. Without this, an agent has no way to decide upfront whether to include discount codes or fulfillment details.

4. **Repeated re-discovery**: Every new checkout session re-surfaces the same platform-level information (extensions, intervention types) that changes infrequently. There is no cacheable pre-session response.

### Why not just use inline capabilities?

Inline capabilities on `POST /checkout_sessions` are the right mechanism for session-level negotiation. They correctly handle:

- **Merchant-specific capabilities**: Payment methods and handlers that vary per merchant.
- **Feature-flagged rollouts**: A merchant might be rolling out Apple Pay at 10% of transactions, so availability can differ between requests for the same merchant.
- **Context-dependent capabilities**: Features that depend on order amount, buyer location, or item type.

[...]

Issue #135, opened 2026-02-12
[SEP][Proposal]: MCP Transport Binding for Agentic Checkout


nvp-stripe:
# SEP: MCP Transport Binding for Agentic Checkout

**Status:** `proposal`
**Type:** Major Change

---

## Abstract

ACP defines a REST API for AI-agent-driven checkout. This works, but it means every agent framework that speaks MCP needs a bespoke adapter to talk to ACP merchants. This SEP proposes adding MCP as a second transport binding for ACP's checkout operations.

The change is purely additive: an OpenRPC schema that maps ACP's 5 checkout operations to MCP tools, a binding document that specifies how REST concepts (headers, path params, status codes) translate to JSON-RPC, and example request/response pairs. No changes to the existing REST spec, JSON Schemas, or protocol semantics.

The goal is to let agents call `create_checkout_session` as a native tool call instead of constructing HTTP requests, while keeping REST as the primary transport for non-agent consumers.

---

## Motivation

ACP is a protocol designed for AI agents. MCP is how AI agents call tools. Today, these two things don't connect without glue code.

Every major agent platform is converging on MCP as the standard for tool integration. When an LLM needs to interact with an external service, it discovers available tools, reads their schemas, and emits structured `tools/call` invocations. The model doesn't think in terms of HTTP methods, URL paths, or request headers. It thinks in terms of function names and arguments.

Right now, to use ACP from an MCP-compatible agent, someone needs to build and maintain an adapter that translates between the two protocols. This is straightforward engineering, but it's also unnecessary friction that every integrator has to solve independently. The translation is mechanical and deterministic - which is exactly why it should be standardized once in the spec rather than reimplemented N times.

ACP's surface area is well-suited for MCP. Five operations, clean request/response schemas, stateless interactions. This is about as simple as an MCP binding gets. There's no reason not to define the canonical mapping and let the ecosystem use it.

---

### Overview

This SEP introduces MCP as a transport binding for ACP's Agentic Checkout API. The binding defines how the 5 REST operations map to MCP tools over JSON-RPC 2.0 using MCP's Streamable HTTP transport.

All domain schemas (request bodies, response bodies, error structures) are reused from the existing JSON Schema bundle at `schema.agentic_checkout.json`. The OpenRPC schema references these via `$ref` - there is no duplication.

### New Files

| File | Purpose |
|---|---|
[...]

PR #77, opened 2026-01-16
[SEP] capability negotiation

prasad-stripe:
# SEP: Capability Negotiation

## 📋 SEP Metadata

- **Author(s)**: Protocol Maintainers
- **Status**: `proposal`
- **Type**: [x] Major Change [ ] Process Change
- **Related Issues/RFCs**: [rfcs/rfc.capability_negotiation.md](rfcs/rfc.capability_negotiation.md)

---

## 🎯 Abstract

This SEP proposes adding **Capability Negotiation** to the Agentic Commerce Protocol using a single `capabilities` object. The same schema is used in requests and responses. Context (request vs response) determines the party. Agents declare capabilities (supported interventions, display context) in requests. Sellers compute and return the **intersection** of supported interventions in responses. The `capabilities` field is **REQUIRED** to ensure predictable behavior. This enables early incompatibility detection, eliminates client-side matching logic, and provides a single source of truth for what will work in each session.

---

### Problems Being Solved

Without capability negotiation, the protocol suffers from:

1. **Failed transactions**: Payment methods are incompatible but only discovered after authorization attempt
2. **Poor UX**: Authentication requirements can't be satisfied by the agent's environment (e.g., 3DS challenge in a non-browser context)
3. **Wasted API calls**: Multiple round-trips to discover incompatibilities
4. **Fragile integrations**: Requires out-of-band coordination between agents and sellers
5. **No optimization**: Sellers can't tailor experiences based on agent capabilities

### Real-World Example

An agent running in a voice-only context receives a checkout session requiring 3D Secure with visual challenge. Without capability negotiation:
- Agent attempts payment → PSP returns 3DS challenge URL
- Agent can't display visual challenge → payment fails
- User experiences friction and abandonment

With capability negotiation:
- Agent declares `supported: []` (no interventions)
- Seller responds with `supported: []`, `required: ["3ds"]` and status `not_ready_for_payment`
- **Intersection is empty** (no common ground)
- Agent knows upfront and can inform user to complete purchase in browser
- Clear, predictable failure with actionable guidance

---

### Schema Changes

**Added `Capabilities` Object:**
```yaml
capabilities:
  type: object
  required: true
  properties:
    payment_methods:
      type: array  # responses only
      items:
        oneOf:
          - type: string  # e.g., "card", "wallet.apple_pay"
          - type: object  # e.g., { method: "card", brands: ["visa"] }
    interventions:
      type: object
      properties:
        supported:
          type: array
[...]

Issue #15, opened 2025-10-02
ACP Manifest and Agent Whitelisting Flow


fboldo:
An essential building block for the Agentic Commerce Protocol would be to standardise a way to express a manifest that defines a store's endpoints, metadata, and more. This addresses critical adoption issues such as endpoint clashes and significantly enhances discoverability for independent agents.

The manifest could be declared using existing web technologies, such as a script tag with a custom type, or a file inside `/.well-known`.

```html
<script type="text/acp+json">
{
    "name": "Example Store",
    "version": "1.0", 
    "description": "An example store for the Agentic Commerce Protocol.",
    "resources": {
        "checkout_sessions": "https://example.com/acp/checkout_sessions",
        "delegate_payment": "https://example.com/acp/delegate_payment"
    }
}
</script>
```

Metadata and resource mapping are both very obvious use cases, but I believe the manifest can play an important role in decentralised agent adoption while also protecting the merchant from malicious usage.

## Agent Whitelisting

The manifest may include a dedicated section indicating which agents are permitted to use the ACP endpoints based on a whitelist of public keys. These keys would be utilised by the agents to sign requests, and an agent would determine whether they are authorised to proceed by referring to the store manifest. With this approach, the `Authorization` header could be entirely replaced by the `Signature` header—the server implementing the protocol could reject any request from non-whitelisted agents.

If the agent key is not listed in the manifest, the agent will be able to request whitelisting by sending a public URL containing their JWKS (JSON Web Token Set) to the merchant, who could use the domain authority (e.g. a JWKS inside the openai.com domain) and/or a human in the loop to approve and whitelist the key in their manifest.

Mechanisms could be implemented to handle agent acknowledgement for the asynchronous approval, such as the agent providing a callback to the request whitelisting endpoint.

### Flow

| Step | Agent Action | Store Action | Manifest Role |
| :--- | :--- | :--- | :--- |
| **1. Discover** | Agent fetches the store's Manifest from the `<script>` tag. | N/A | Manifest provides the `agent_registration_endpoint`. |
| **2. Request** | Agent sends a POST request to the `agent_registration_endpoint` with its desired JWKS URL and identifying metadata (e.g., agent name, contact info). | Store receives the request and initiates an internal approval process. | Manifest declares where the request must be sent. |
[...]

PR #137, opened 2026-02-12
SEP: Discovery Well-Known Document

nvp-stripe:
# SEP: Discovery Well-Known Document

## 📋 SEP Metadata

- **SEP Number**: #136
- **Author(s)**: [Your Name]
- **Related Issues/RFCs**: [#15](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/15), [#135 (MCP Transport Binding)](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/135), `rfcs/rfc.capability_negotiation.md`

---

## 🎯 Abstract

Agents currently have no way to check whether a platform supports ACP without creating a checkout session. This SEP adds a well-known discovery document at `/.well-known/acp.json` (per RFC 8615) that returns platform-level metadata including protocol version, API base URL, supported transports, available services, extensions, intervention types, currencies, and locales. The document is static, publicly accessible, and does not require authentication.

The document is intentionally scoped to platform-level information that is stable across all merchants. Merchant-specific and session-specific details (payment methods, payment handlers, feature-flagged capabilities) continue to be negotiated inline via the `capabilities` object on `POST /checkout_sessions`.

---

## 💡 Motivation

Without a discovery mechanism, agents face several problems:

- **Blind first requests**: Agents have to attempt `POST /checkout_sessions` and interpret failure responses to figure out if a platform supports ACP at all.
- **Version guessing**: No way to learn the supported API version before making a request. The first call may fail with a version mismatch error.
- **No base URL bootstrapping**: Agents need the full API base URL to make any ACP call. Without discovery, the base URL has to be communicated entirely out-of-band.
- **No feature overview**: Agents can't determine which extensions (discounts, fulfillment) or services (orders, delegate payment) are available without starting a transaction.
- **No transport awareness**: Agents can't determine whether a platform supports MCP in addition to REST without trial and error.
- **Repeated re-discovery**: Every new session resurfaces the same platform-level information that changes infrequently, with no way to cache it.

The capability negotiation SEP (#77) explicitly listed "Creating a capability discovery service or registry" as a non-goal, deferring it to future work. This SEP picks up that deferred work.

---

### Discovery Document

`/.well-known/acp.json` served at the origin root per RFC 8615. No authentication required. Implementations SHOULD include `Cache-Control: public, max-age=3600` as a recommended minimum.

### New Schemas

**DiscoveryResponse** (top-level document):

| Field | Type | Required | Description |
|---|---|---|---|
| `protocol` | `DiscoveryProtocol` | Yes | Protocol name, version, supported versions, docs URL |
| `api_base_url` | `string` (URI) | Yes | Base URL for the ACP REST API |
| `transports` | `string[]` | Yes | Supported transport bindings (`rest`, `mcp`) |
| `capabilities` | `DiscoveryCapabilities` | Yes | Platform-level capabilities |

**DiscoveryCapabilities** (wraps feature declarations inside `capabilities`, following the same pattern used in checkout session requests/responses):

| Field | Type | Required | Description |
|---|---|---|---|
| `services` | `string[]` | Yes | ACP services implemented (closed enum: `checkout`, `orders`, `delegate_payment`) |
| `extensions` | `DiscoveryExtension[]` | No | Extensions the platform supports |
| `intervention_types` | `string[]` | No | Available intervention types (closed enum) |
| `supported_currencies` | `string[]` | No | ISO 4217 currency codes |
| `supported_locales` | `string[]` | No | BCP 47 locale tags |

**DiscoveryProtocol**: `name` (always `"acp"`), `version`, `supported_versions` array (chronological order, oldest first), optional `documentation_url`.

**DiscoveryExtension**: Lighter than `ExtensionDeclaration`. Just `name` and optional `spec` URL. Omits session-level fields like `extends` and `schema`.

### Example Response

```json
{
  "protocol": {
    "name": "acp",
    "version": "2026-01-30",
    "supported_versions": ["2025-09-29", "2025-12-12", "2026-01-16", "2026-01-30"],
    "documentation_url": "https://agenticcommerce.dev"
  },
  "api_base_url": "https://acp.stripe.com/api",
  "transports": ["rest", "mcp"],
  "capabilities": {
    "services": ["checkout", "orders", "delegate_payment"],
    "extensions": [
      { "name": "discount", "spec": "https://agenticcommerce.dev/specs/discount" },
      { "name": "fulfillment", "spec": "https://agenticcommerce.dev/specs/fulfillment" }
    ],
    "intervention_types": ["3ds", "biometric", "address_verification"],
    "supported_currencies": ["usd", "eur", "gbp"],
    "supported_locales": ["en-US", "fr-FR", "de-DE"]
  }
}
```

### Files Changed

- `spec/unreleased/json-schema/schema.agentic_checkout.json` - `DiscoveryResponse`, `DiscoveryCapabilities`, `DiscoveryProtocol`, `DiscoveryExtension` in `$defs`
- `rfcs/rfc.discovery.md` - Full RFC
- `examples/unreleased/examples.agentic_checkout.json` - Full and minimal discovery response examples
- `changelog/unreleased/discovery-well-known.md` - Changelog entry
- `docs/mcp-binding.md` - Updated Discovery section to reference `/.well-known/acp.json` and `transports` field

Note: the OpenAPI spec (`openapi.agentic_checkout.yaml`) is intentionally **not** modified. The well-known document is a static file served outside the API surface, so it does not belong in the OpenAPI definition.

---

## 🤔 Rationale

**`.well-known` instead of a dedicated API endpoint**: Discovery solves a bootstrapping problem. Agents need to learn the API base URL before they can call any endpoint. A `/.well-known/acp.json` document at the origin root means agents only need a domain to get started. This follows established precedent from OpenID Connect (`/.well-known/openid-configuration`), OAuth 2.0 Authorization Server Metadata, and others. A dedicated `GET /capabilities` endpoint would require agents to already know the API base URL, which defeats the purpose of discovery.

**Platform-level, not merchant-level**: ACP is deployed through intermediaries serving many merchants behind a single base URL. Merchant identity comes from the Bearer token, which isn't available for an unauthenticated document. Merchant capabilities are also subject to feature flags and gradual rollouts, so they can't be represented deterministically in a static document.

**Capabilities wrapper**: Extensions, services, and intervention types are nested inside a `capabilities` object, matching the pattern used by `POST /checkout_sessions` request/response bodies.

**No payment methods**: Payment method availability is merchant-specific, feature-flagged, and context-dependent. A merchant might be rolling out a new payment method at 10% of transactions. Including them would be either inaccurate or misleading.

**Closed enums with versioning**: `services`, `intervention_types`, and `transports` use closed enums. New values are introduced in new API versions. Agents can't do anything useful with a service name they don't have a schema for, so open enums would just be noise. Closed enums per version keep things predictable.

**`supported_versions` ordering**: The array is specified as chronological (oldest first). This avoids agents needing to parse and sort date strings to find the latest mutual version.

**No `merchant_id`**: The document is unauthenticated. Accepting merchant identifiers would allow anyone to enumerate which merchants exist on a platform, creating fingerprinting and enumeration risks. This is documented as a non-goal.

**Cache-Control guidance**: A recommended minimum of `Cache-Control: public, max-age=3600` is specified. Platform capabilities don't change frequently. Without guidance, implementations would either skip caching (unnecessary load) or cache too aggressively (stale data after version upgrades).

**`transports` field**: Cross-references SEP #135 (MCP Transport Binding). When a platform supports MCP, the `transports` array includes `"mcp"`, letting agents discover transport options alongside capabilities in a single request.

---

## 🔄 Backward Compatibility

No breaking changes. This is purely additive:

- New well-known document (`/.well-known/acp.json`) that doesn't conflict with existing endpoints or paths
- New schemas that don't modify existing schemas
- `POST /checkout_sessions` and inline capability negotiation are completely unchanged
- Agents that don't use discovery continue working as before

---

RFC, JSON Schema, examples, and changelog are included in this PR. The discovery schemas are defined in `schema.agentic_checkout.json` and validate against draft 2020-12.

---

## 🔒 Security Implications

- **No sensitive data**: The document contains only public platform metadata. No merchant identifiers, payment handler configs, PSP routing, or buyer info.
- **Rate limiting recommended**: The document is publicly accessible, so implementations should apply rate limiting and return `429` with `Retry-After` headers.
- **Merchant enumeration prevention**: The document is platform-scoped and must not accept or return merchant identifiers. This is documented as a non-goal with explicit rationale.
- **Minimal information disclosure**: Comparable to publishing an OpenAPI spec. Does not enable merchant fingerprinting.

---

## ✅ Pre-Submission Checklist

- [x] I have signed the [Contributor License Agreement (CLA)](legal/cla/INDIVIDUAL.md)
- [x] This PR includes a changelog entry file in `changelog/unreleased/discovery-well-known.md`

---

## 📚 Additional Context

Related to [#15 (ACP Manifest and Agent Whitelisting Flow)](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/15) but narrower in scope. Issue #15 covers endpoint resolution and agent whitelisting. This SEP only covers capability discovery. The two are complementary.

Complementary to [#135 (MCP Transport Binding)](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/135). Discovery tells agents what's available and which transports are supported; the MCP binding tells them how to use MCP specifically. The `transports` field in the discovery document provides the connection point between the two.

---

## 🙋 Questions for Reviewers

1. Is `api_base_url` the right field name, or should it be something like `endpoint` or `base_url`?
2. Should the spec mandate that `transports` always includes `"rest"`, or could a platform theoretically be MCP-only?
3. Any objections to recommending `Cache-Control: public, max-age=3600` as the minimum? Happy to adjust the floor.

### DIFF
--- a/examples/unreleased/examples.agentic_checkout.json
+++ b/examples/unreleased/examples.agentic_checkout.json
@@ -1484,11 +1484,13 @@
       "extensions": [
         {
           "name": "discount",
-          "spec": "https://agenticcommerce.dev/specs/discount"
+          "spec": "https://agenticcommerce.dev/specs/discount",
+          "schema": "https://agenticcommerce.dev/schemas/discount.json"
         },
         {
           "name": "fulfillment",
-          "spec": "https://agenticcommerce.dev/specs/fulfillment"
+          "spec": "https://agenticcommerce.dev/specs/fulfillment",
+          "schema": "https://agenticcommerce.dev/schemas/fulfillment.json"
         }
       ],
       "intervention_types": [
--- a/rfcs/rfc.discovery.md
+++ b/rfcs/rfc.discovery.md
@@ -137,6 +137,7 @@ The document is a `DiscoveryResponse` object containing the following fields:
 |---|---|---|---|
 | `name` | `string` | Yes | Extension identifier (e.g., `"discount"`, `"fulfillment"`). |
 | `spec` | `string` (URI) | No | URL to the extension's specification document. |
+| `schema` | `string` (URI) | No | URL to the extension's JSON Schema definition for programmatic validation. |
 
 #### Services Enum Values
 
@@ -212,8 +213,8 @@ Cache-Control: public, max-age=3600
   "capabilities": {
     "services": ["checkout", "orders", "delegate_payment"],
     "extensions": [
-      { "name": "discount", "spec": "https://agenticcommerce.dev/specs/discount" },
-      { "name": "fulfillment", "spec": "https://agenticcommerce.dev/specs/fulfillment" }
+      { "name": "discount", "spec": "https://agenticcommerce.dev/specs/discount", "schema": "https://agenticcommerce.dev/schemas/discount.json" },
+      { "name": "fulfillment", "spec": "https://agenticcommerce.dev/specs/fulfillment", "schema": "https://agenticcommerce.dev/schemas/fulfillment.json" }
     ],
     "intervention_types": ["3ds", "biometric", "address_verification"],
     "supported_currencies": ["usd", "eur", "gbp"],
@@ -357,7 +358,8 @@ This RFC provides a foundation for future discovery enhancements:
 - **Webhook capabilities**: Advertising supported webhook event types and delivery mechanisms.
 - **Authentication methods**: Declaring supported authentication mechanisms (e.g., OAuth 2.0 identity linking) when those capabilities are added to ACP.
 - **Service-level metadata**: Adding per-service configuration (e.g., maximum line items, supported fulfillment types) as the platform's feature set grows.
-- **MCP endpoint URL**: Adding an explicit `mcp_url` field when the MCP transport binding is finalized, allowing agents to discover the MCP endpoint alongside the REST base URL.
+- **Transport endpoint discovery**: Structured transport objects with per-transport endpoint URLs (e.g., `{"type": "rest", "url": "..."}`, `{"type": "mcp", "url": "..."}`), enabling agents to discover MCP and other transport endpoints directly from the discovery document. Deferred pending MCP binding finalization (SEP #135).
+- **Signing keys**: Public key advertisement (JWK format) for signature verification, enabling agents to verify the authenticity of responses. Deferred pending formalization of ACP's request signing specification.
 
 ---
 
--- a/spec/unreleased/json-schema/schema.agentic_checkout.json
+++ b/spec/unreleased/json-schema/schema.agentic_checkout.json
@@ -3751,7 +3751,7 @@
       "required": ["services"],
       "example": {
         "services": ["checkout", "orders", "delegate_payment"],
-        "extensions": [{ "name": "discount", "spec": "https://agenticcommerce.dev/specs/discount" }],
+        "extensions": [{ "name": "discount", "spec": "https://agenticcommerce.dev/specs/discount", "schema": "https://agenticcommerce.dev/schemas/discount.json" }],
         "intervention_types": ["3ds"],
         "supported_currencies": ["usd", "eur"],
         "supported_locales": ["en-US"]
@@ -3809,12 +3809,18 @@
           "type": "string",
           "format": "uri",
           "description": "URL to the extension's specification document."
+        },
+        "schema": {
+          "type": "string",
+          "format": "uri",
+          "description": "URL to the extension's JSON Schema definition for programmatic validation."
         }
       },
       "required": ["name"],
       "example": {
         "name": "discount",
-        "spec": "https://agenticcommerce.dev/specs/discount"
+        "spec": "https://agenticcommerce.dev/specs/discount",
+        "schema": "https://agenticcommerce.dev/schemas/discount.json"
       }
     },
     "Error": {


aravindrao-openai:
Thanks for the PR! I'd like to explore what gap a `capabilities` endpoint is filling that justifies not supporting `.well_known` which is standard across other protocols. My main concern is that without the ability to advertise the base URL through `.well_known` (which is required to hit the `capabilities` endpoint), there will be issues with service discovery.

kxbnb:
@aravindrao-openai The PR already proposes `/.well-known/acp.json`, not a capabilities endpoint — so your bootstrap concern is actually addressed here. An agent only needs the domain to fetch the well-known file, and `api_base_url` in the response tells them where to send API calls.

I argued for an endpoint over well-known in #136, mainly around dynamic fields and cache invalidation. But the bootstrap argument is real — agents need to discover the base URL before they can call any endpoint, and well-known solves that without out-of-band coordination. That tips the balance.

On your inline comment about the `capabilities` wrapper — the PR description already nests `services`, `extensions`, and `intervention_types` under a `capabilities` object in the response. If the implementation doesn't match that yet, +1 on aligning it.

One thing that might help both of you: there's a middle ground where well-known handles the bootstrap (protocol version, base URL, transports) and the richer capabilities data is fetched from an authenticated endpoint. That way the well-known file stays genuinely static and platform-scoped, while merchant-specific or feature-flagged capabilities (which will come eventually) don't need to be shoehorned into a public document. But that's probably a v2 concern.

prasad-stripe (approved):
Thanks for incorporating the feedback!

PR #139, opened 2026-02-13
SEP: MCP Transport Binding for Agentic Checkout

nvp-stripe:
# SEP: MCP Transport Binding for Agentic Checkout

## 📋 SEP Metadata

- **SEP Number**: #135
- **Status**: `proposal` 
- **Related Issues/RFCs**: #135

---

## 🎯 Abstract

ACP currently supports REST as its only transport. This SEP adds MCP (Model Context Protocol) as a second transport binding, enabling AI agents to invoke ACP's 5 checkout operations as native MCP tool calls over JSON-RPC 2.0.

The change is purely additive: an OpenRPC schema maps ACP's REST operations to MCP tools, a binding document specifies how REST concepts (headers, path params, status codes) translate to JSON-RPC, and example request/response pairs demonstrate the wire format. No changes to the existing REST spec, JSON Schemas, or protocol semantics.

---

## 💡 Motivation

ACP is designed for AI agents. MCP is how AI agents call tools. Today these don't connect without a bespoke adapter that every integrator must build independently.

Every major agent platform is converging on MCP as the standard for tool integration. The translation from ACP REST to MCP is mechanical and deterministic -- it should be standardized once in the spec rather than reimplemented N times.

ACP's 5-operation surface area is well-suited for MCP. There's no reason not to define the canonical mapping and let the ecosystem use it.

---

## 📐 Specification

This PR adds 4 new files (no existing files modified):

| File | Purpose |
|---|---|
| `spec/unreleased/openrpc/openrpc.agentic_checkout.json` | OpenRPC 1.3.2 schema defining 5 MCP tools |
| `docs/mcp-binding.md` | Binding specification: mapping conventions, error handling, conformance |
| `examples/unreleased/examples.mcp.agentic_checkout.json` | JSON-RPC request/response examples for all tools + error cases |
| `changelog/unreleased/mcp-transport-binding.md` | Changelog entry |

### Tools

| MCP Tool | REST Operation |
|---|---|
| `create_checkout_session` | `POST /checkout_sessions` |
| `get_checkout_session` | `GET /checkout_sessions/{id}` |
| `update_checkout_session` | `POST /checkout_sessions/{id}` |
| `complete_checkout_session` | `POST /checkout_sessions/{id}/complete` |
| `cancel_checkout_session` | `POST /checkout_sessions/{id}/cancel` |

### Argument Structure

Every tool uses a consistent `meta` / `id` / `payload` structure:

- **`meta`**: Protocol metadata mapped from [ACP HTTP headers](https://www.agenticcommerce.dev/docs/reference/checkout#required-headers) (`api_version`, `idempotency_key`, etc.). Auth is out-of-band via MCP server configuration.
- **`id`**: Resource identifier mapped from the path parameter (omitted for create).
- **`payload`**: Domain data mapped from the REST request body. `$ref`s existing ACP request schemas directly (`CheckoutSessionCreateRequest`, etc.) -- no duplication.

### Error Handling

All ACP errors use JSON-RPC error code `-32000` uniformly. The ACP `Error` object (`type`, `code`, `message`, `param`) is placed in the JSON-RPC `error.data` field where consumers should inspect it.

Full details in `docs/mcp-binding.md`.

---

## 🤔 Rationale

- **Spec-only, no runtime**: The binding is a thin translation layer. A reference MCP server implementation can follow once the binding is stable.
- **5 tools, 1:1 mapping**: LLMs reason better about tools with distinct purposes than a single tool with an `action` param.
- **`payload` wrapper**: Enables direct `$ref` to existing ACP request schemas without composition. Clean separation from `meta` (transport) and `id` (routing).
- **Auth out-of-band**: Bearer tokens in tool arguments would be visible in LLM context windows. MCP handles auth at the server level.
- **Tools, not Resources**: Checkout sessions are transient, agent-driven objects. MCP Resources are better suited for future capabilities (product catalog, order history) where subscriptions and URI-based access make sense.
- **No discovery profile**: ACP's inline capability negotiation means no `/.well-known/` endpoint is needed.
- **Checkout scope only**: Delegate Payment is a separate API surface served by a different party and should be a follow-up SEP.

---

## 🔄 Backward Compatibility

No breaking changes. Purely additive.

- REST API, JSON Schemas, and all existing protocol behavior are unchanged.
- Merchants that don't want MCP support ignore the OpenRPC schema entirely.
- No new required fields on any existing schema.
- Existing tooling (OpenAPI code generators, validation scripts) is unaffected.
- The OpenRPC schema lives in a new `spec/unreleased/openrpc/` directory with no interaction with existing directories beyond `$ref` links.

---

Not included in this initial proposal. A reference implementation (likely a TypeScript MCP server that proxies to any ACP REST endpoint) will be developed if the SEP is accepted. The server would be a thin, table-driven adapter that translates `tools/call` to REST `fetch()` calls.

---

## 🔒 Security Implications

- **Auth tokens excluded from tool arguments** by design. MCP tool schemas are visible to the LLM during planning; keeping auth at the server configuration level avoids token exposure.
- **Request signing preserved**: `Signature` and `Timestamp` map to `meta.signature` and `meta.timestamp`.
- **Idempotency preserved**: `Idempotency-Key` maps to `meta.idempotency_key`.
- **No new attack surface** beyond the MCP endpoint itself, which should enforce the same authorization, rate limiting, and input validation as the REST endpoint.

---
- ## 📚 Additional Context

---

## ✅ Pre-Submission Checklist

- [x] I have signed the [Contributor License Agreement (CLA)](../legal/cla/INDIVIDUAL.md)
- [x] This PR includes a changelog entry file in `changelog/unreleased/mcp-transport-binding.md`

---

## 🙋 Questions for Reviewers

1. **`meta` naming**: The `meta` object is a convention we're using, not an MCP standard. Should it be called something else (`protocol`, `headers`, `_meta`)?
2. **Error code mapping**: We use `-32000` uniformly for all ACP errors with semantics in `data`. Should we instead map HTTP status categories to specific JSON-RPC codes?
3. **Delegate Payment**: Should we fast-track a companion SEP for the Delegate Payment MCP binding, or let this land first?

### DIFF
--- a/docs/mcp-binding.md
+++ b/docs/mcp-binding.md
@@ -14,7 +14,9 @@ protocol semantics are unchanged.
 All domain schemas are reused from the existing JSON Schema bundle
 (`schema.agentic_checkout.json`). The OpenRPC schema at
 `spec/unreleased/openrpc/openrpc.agentic_checkout.json` references these via
-`$ref`. No domain types are duplicated.
+`$ref`. No domain types are duplicated. These references use relative file
+paths for spec-time validation. Implementers **MUST** resolve and bundle all
+`$ref` targets before serving tool schemas to MCP clients.
 
 ### Transport
 
@@ -31,6 +33,12 @@ merchant responds with the negotiated set), so there is no pre-session
 discovery phase. The MCP endpoint URL is communicated out-of-band, the same
 way REST endpoint URLs are communicated today (documentation, onboarding, API
 configuration).
+
+The required `meta.api_version` field presumes the agent knows a compatible
+API version before its first tool call. Agents may obtain this from merchant
+documentation, or via the Discovery Capabilities mechanism
+([SEP #135](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/135))
+once available.
 
 ## Tool Definitions
 
@@ -67,6 +75,9 @@ Every tool call uses a consistent three-field argument structure:
 | `meta`    | Protocol metadata (from HTTP headers)| All tools (required)                          |
 | `id`      | Resource identifier (from path param)| get, update, complete, cancel (required)      |
 | `payload` | Domain data (from request body)      | create, update, complete (required); cancel (optional); get (absent) |
+
+When `payload` is marked optional (e.g., `cancel_checkout_session`), omitting
+the field entirely is valid. Servers **MUST NOT** require `payload: null`.
 
 ### Why `payload`?
 
@@ -106,6 +117,9 @@ handle authentication at the connection level (via server configuration or
 OAuth), not per tool call. Including bearer tokens in tool arguments would
 expose them in tool schemas visible to LLMs.
 
+The `meta` object is extensible (`additionalProperties: true` in the OpenRPC
+schema). Servers **SHOULD** ignore unrecognized `meta` fields to allow
+forward-compatible header additions without requiring a schema version bump.
 If future ACP versions introduce new protocol-level headers, corresponding
 `meta` fields SHOULD be added to the MCP binding.
 
@@ -154,10 +168,13 @@ schema (`type`, `code`, `message`, `param`) is placed in the JSON-RPC error's
 }
 ```
 
-All ACP errors use JSON-RPC error code `-32000` (server error). The useful
-error semantics are in ACP's `Error` object in `data` — consumers should
-inspect `data.type` and `data.code` to understand the error, not the JSON-RPC
-error code.
+All ACP errors use JSON-RPC error code `-32000` (server error) uniformly. The
+useful error semantics are in ACP's `Error` object in `data` -- consumers
+**MUST** inspect `data.type` and `data.code` to understand the error, not the
+JSON-RPC error code. Clients that display or log JSON-RPC error codes without
+inspecting `data` will see a uniform `-32000` for all ACP errors. This is a
+known tradeoff: the JSON-RPC error code layer is intentionally thin, and the
+ACP error model in `data` carries the actionable semantics.
 
 The exception is `-32602` (Invalid params), which is reserved for malformed
 requests that fail at the JSON-RPC level before reaching ACP business logic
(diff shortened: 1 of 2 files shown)


nvp-stripe:
Thanks for the thorough reviews everyone. Pushed an update that addresses all the feedback. Here's what changed and why:

**`meta` extensibility (`additionalProperties: true`)**

This was the most important fix. Flipped `additionalProperties` from `false` to `true` on the meta schema in the OpenRPC spec. The previous version would have silently rejected any new header fields (like a future `Trace-Id` for distributed tracing) until we shipped a schema update. That's unnecessary friction for what's essentially a transport envelope. Added a note to the binding doc clarifying that servers SHOULD ignore unreocgnized meta fields.

**Auth error types (401/403)**

Looking at this closely, the gap is actually in ACP's core `Error` schema, not just the MCP binding. The `Error.type` enum only has `invalid_request`, `processing_error`, and `service_unavailable` -- there's no `authentication_error` or `authorization_error`. Rather than inventing auth error types only for the MCP binding (which would create a divergence between transports), I've documented the current behavior: 401/403 from the merchant REST API surface as `-32000` with `data.type: "invalid_request"` until ACP adds dedicated auth error types. Added an inline example showing how a 403 propagates. The right fix is adding auth types to the core Error schema, but thats a separate change that benefits both transports.

Also added a note that MCP server-level auth failures (bad OAuth token to the MCP endpoint itself) are outside the ACP error model entirely and should use standard MCP error handling.

**`$ref` resolution**

Added a one-liner in the Overview noting that the relative `$ref` paths are for spec-time validation and implementers MUST resolve/bundle them before serving tool schemas to clients.

**Uniform `-32000` tradeoff**

Kept the uniform `-32000` design but expanded the error handling section to call out the tradeoff explicitly. Consumers MUST inspect `data.type` and `data.code` -- if they only look at the JSON-RPC error code they'll see the same thing for every error. I think this is the right call, the useful signal lives in ACP's error model not in JSON-RPC codes, but it should be documented rather then left implicit.

**`cancel_checkout_session` payload -- absent vs null**

Clarified in the Argument Structure section that when payload is optional, omitting the key entirely is valid. Servers MUST NOT require `payload: null`. So `{ "meta": {...}, "id": "..." }` is a valid cancel call.

**Scope -- "Part 1 of N"**

Added explicit language that this is the first MCP binding for ACP and a follow-up SEP will cover Delegate Payment to complete the full agent checkout flow over MCP.

**Proxy trust boundary / PCI**

Good catch from the review on #135. Added a Security Considerations section to the binding doc. When an MCP server runs as a proxy, it sees everything including payment tokens in `complete_checkout_session`. Proxy operators should evaluate PCI scope. Co-locating with the REST backend is the cleanest path since it avoids introducing a new intermediary.

**Version bootstrapping**

Added a cross-reference in the Discovery section noting that `meta.api_version` presumes the agent knows a compatible version. Agents can get this from merchant docs or via the Discovery Capabilities mechanism (SEP #135) once that's available. This isn't a new problem -- REST has the same blind-first-request issue -- but worth calling out the connection.

**No changes for: Resource vs Tool for get_checkout_session**

I looked at this again and still think all-Tools is correct. Checkout sessions are transient and agent-driven, the agent knows when state changed because it just caused the change. Resources would add complexity for a polling pattern that's already natural. If ACP ever adds server-pushed status changes or product catalogs, Resources would be worth revisiting then.

Issue #142, opened 2026-02-15
[Feature Request] Add redirect/webview intervention types for wallet handlers


kxbnb:
**Problem**

The current `intervention_types` enum covers `3ds`, `biometric`, and `address_verification`. These are all inline flows — the agent can handle them within the checkout session context.

Wallet handlers like Stripe Link (#141), and eventually Apple Pay and Google Pay, require interactive flows that don't fit this model. Link uses email + OTP authentication — the buyer needs to complete an out-of-band step (check email, enter code) before the agent can proceed. The agent has no standard way to surface this to the buyer.

**Example flow that breaks today**

1. Agent calls `complete_checkout_session` with a Link payment handler
2. Link requires email + OTP authentication
3. Agent gets... what? There's no intervention type to describe "hand control to the buyer for an interactive auth flow"
4. Agent can't communicate to the buyer what's needed

**Proposed addition**

Add one or more of:
- `redirect` — buyer needs to complete a flow at an external URL (OAuth-style)
- `webview` — buyer needs to interact with an embedded UI (SDK-rendered)
- `otp` — buyer needs to enter a one-time code sent via email/SMS

These cover the main interactive patterns across wallet providers:
- Stripe Link: `otp` (email code)
- Apple Pay: `biometric` (already covered) or `webview` (device prompt)
- Google Pay: `redirect` or `webview` (Google account auth)
- Shop Pay: `otp` (SMS code)

**Why this matters now**

SEP #141 (Stripe Link handler) is the first wallet handler being specified. The pattern it sets will get copied for Apple Pay, Google Pay, and Shop Pay. If we don't sort out intervention types before the first handler lands, each handler will work around the gap differently, and we'll end up with inconsistent approaches.

**Relationship to other SEPs**

- #141 (Stripe Link) — needs this for the email + OTP flow
- #137 (Discovery Capabilities) — `intervention_types` in the discovery response should include these new types
- #135 / #139 (MCP Transport Binding) — the MCP binding error model needs to handle intervention handoffs

Happy to draft a SEP if there's interest in formalizing this.

kxbnb:
Getting this sorted before #141 lands is the right call — whatever pattern Stripe Link sets will get copied for every wallet handler after it. Retrofitting is always harder.

[...]

Issue #141, opened 2026-02-13
[SEP] Stripe Link Payment Handler


prasad-stripe:
# SEP: Stripe Link Payment Handler (`com.stripe.link`)

## SEP Metadata

- **Author(s)**: Prasad Wangikar / Stripe
- **Status**: `proposal`
- **Type**: [x] Major Change [ ] Process Change
- **Related Issues/RFCs**: `rfcs/rfc.payment_handlers.md` (payment handlers framework), `rfcs/rfc.delegate_payment.md` (delegate payment API)

---

## Abstract

This SEP introduces `com.stripe.link` as a new payment handler in the ACP payment handlers framework. Link is Stripe's accelerated checkout wallet that lets buyers save and reuse payment methods across any Link-enabled merchant. The handler enables a flow where agents initiate Link, the buyer authenticates and selects a saved payment method, Link returns a token, and the agent delegates the token via `delegate_payment` to obtain an SPT before completing checkout. The seller never sees raw payment credentials.

This is the first **accelerated-checkout handler** in ACP — a handler where the buyer authenticates directly with a third-party wallet rather than providing credentials to the agent. It sets a pattern for future wallet-type handlers (e.g., Shop Pay, Apple Pay, Google Pay) in ACP.

---

### Why add Link to ACP?

1. **Higher conversion**: Link's 200M+ consumer network and one-click checkout experience increases conversion for sellers in agentic commerce, just as it does in traditional web checkout.

2. **No credential handling by agents**: Agents never see or handle raw payment credentials. The buyer authenticates with Link directly; the agent receives only a scoped token.

3. **Accelerated-checkout pattern**: ACP currently defines only `dev.acp.tokenized.*` handlers where the agent collects raw credentials. Link introduces a different acquisition pattern — the wallet handles credential collection and authentication, returning a token directly. The agent then delegates this token through the standard `delegate_payment` flow like any other handler.

4. **Already referenced in ACP**: The payment handlers RFC (section 9.1, migration example) already lists `com.stripe.link` as a handler name. This SEP formalizes its specification.

### Why the existing framework needs this handler

The `dev.acp.tokenized.*` handlers assume the agent collects credentials directly from the buyer. Link's credential acquisition is different:

1. Agent initiates Link with merchant identity
2. **Buyer authenticates directly with Link** (email + OTP)
3. Link returns a token bound to the checkout and merchant
[...]

Issue #98, opened 2026-02-03
[Documentation] Document `intervention_required` error code


dean-harel-nekuda-ai:
## Summary

The RFC documents message error codes (`missing`, `invalid`, `out_of_stock`, `payment_declined`, `requires_sign_in`, `requires_3ds`) but capability mismatch scenarios use `intervention_required`, which is **not in the documented list**. This code should be formally documented.

## Motivation

RFC Section 5.3 (or equivalent) documents these error codes for messages:

```
missing
invalid
out_of_stock
payment_declined
requires_sign_in
requires_3ds
```

However, when a required intervention cannot be satisfied by the agent's capabilities, implementations return `intervention_required`:

```json
{
  "status": "not_ready_for_payment",
  "capabilities": {
    "interventions": {
      "supported": [],
      "required": ["3ds"],
      "enforcement": "always"
    }
  },
  "messages": [
    {
      "type": "error",
      "code": "intervention_required",
      "content_type": "plain",
      "content": "This purchase requires 3D Secure authentication, which cannot be completed in your current environment."
    }
  ]
}
```

This code is referenced in capability negotiation examples but not formally documented in the error codes list.

### Ambiguity with `requires_3ds`

The relationship between `intervention_required` and `requires_3ds` is unclear:

- When should a server return `requires_3ds` vs `intervention_required`?
- Is `requires_3ds` for "please do 3DS" and `intervention_required` for "cannot proceed"?
- Can `intervention_required` apply to non-3DS interventions (biometric, address_verification)?

## Proposal

Add `intervention_required` to the documented error codes:

> **`intervention_required`**: A required intervention (from `capabilities.interventions.required`) cannot be satisfied by the agent's declared capabilities. The checkout cannot proceed in the current environment. The agent should inform the user that this purchase requires capabilities not available in their current session.

### Clarify Relationship to Other Codes

Add guidance distinguishing related codes:

| Code | When to Use |
|------|-------------|
| `requires_3ds` | 3DS is required AND the agent supports it. The agent should initiate the 3DS flow. |
| `intervention_required` | A required intervention is NOT supported by the agent. This is a terminal blocker—the agent should inform the user and cannot proceed. |
| `requires_sign_in` | User authentication is required before checkout can continue. |

### Key Distinction

- `requires_3ds` = "Do this" (actionable by agent)
- `intervention_required` = "Cannot do this" (terminal blocker)

## Implementation Notes

[...]

PR #150, opened 2026-02-17
SEP: Intervention types extension for wallet handler auth flows

kxbnb:
# SEP: Intervention Types Extension

## 📋 SEP Metadata

- **SEP Number**: #142
- **Author(s)**: Karthik / kxbnb
- **Related Issues/RFCs**: #142 (feature request), #141 (Stripe Link handler), #137 (Discovery Capabilities), #98 (intervention_required error), RFC: Capability Negotiation

---

## 🎯 Abstract

The current `intervention_types` enum covers `3ds`, `biometric`, and `address_verification`. These are all inline flows. Wallet handlers like Stripe Link, Google Pay, and Shop Pay require interactive auth flows that don't fit: email/SMS OTP, redirect-based login, and embedded webview auth.

This SEP adds three new intervention types (`redirect`, `webview`, `otp`), introduces `InterventionContext` so agents know what to do when they get an `intervention_required` error, and adds `InterventionResponse` so agents can submit OTP codes and redirect completion signals via the existing update endpoint.

---

## 💡 Motivation

SEP #141 (Stripe Link) is the first wallet handler being specified. Link requires email + OTP authentication. There's currently no way to express "this checkout requires the buyer to enter a one-time code" in the intervention vocabulary.

Without these types, each wallet handler will work around the gap differently. Stripe Link might stuff OTP info into the error message text. Google Pay might invent a custom field for redirect URLs. By the time we have four wallet handlers, we'd have four incompatible approaches.

The `intervention_required` error code already exists in the schema but has no structured payload -- the agent gets the error but not the intervention type, redirect URL, or challenge details. This makes it hard for agents to handle interventions programmatically.

---

## 📐 Specification

See `rfcs/rfc.intervention_types.md` for the full spec. The changes:

**Schema additions:**
- `redirect`, `webview`, `otp` added to `InterventionCapabilities.supported` and `.required` enums
- `InterventionContext` type with `intervention_type`, `redirect_url`, `callback_url`, `challenge_hint`, `expires_at`, `provider`
- `InterventionResponse` type with `type` (otp/redirect_complete/webview_complete), `code`, `callback_data`
- `intervention_context` field on `CheckoutSessionBase`
- `intervention_response` field on `CheckoutSessionUpdateRequest`

**Examples:**
- OTP flow (Stripe Link style): create → intervention_required with context → update with code → ready_for_payment
- Redirect flow (Google Pay style): create → intervention_required with redirect URL → update with completion signal
- Capability mismatch: agent doesn't support required intervention type

---

## 🤔 Rationale

**Mechanism-based vs provider-specific types:** We went with mechanism-based (`redirect`, `webview`, `otp`) over provider-specific (`stripe_link_auth`, `google_pay_auth`). New providers map to existing types instead of requiring enum additions each time. Stripe Link and Shop Pay both use `otp`, so one capability covers both.

**`redirect` separate from `webview`:** Different agent capabilities. A CLI agent can open an external browser for a redirect but can't render an iframe. These need separate capability declarations.

**Session-level `InterventionContext` instead of extending `MessageError`:** Intervention state persists across API calls while the buyer completes the flow. Tying it to a single error message creates lifecycle issues. Session-level placement matches how `authentication_metadata` already works for 3DS.

**`address_verification` stays out of `required`:** It's a passive check, not an interactive flow. Including it in `required` would imply buyer action.

---

## 🔄 Backward Compatibility

All changes are additive -- nothing breaks.

- New enum values in existing arrays. Agents/sellers that don't use them just omit them.
- New optional fields on existing objects. Existing implementations ignore them.
- Negotiation mechanism unchanged.
- Existing types (`3ds`, `biometric`, `address_verification`) work identically.

---

Not yet available. This PR includes the RFC, schema changes, and examples.

---

## 🔒 Security Implications

- **Redirect URLs:** Agents must validate redirect URLs are HTTPS and match known payment providers. Arbitrary redirects could be used for phishing.
- **OTP handling:** Agents must not log or persist codes. Codes go only through `intervention_response.code`, never in free-text fields.
- **Challenge hints:** Must be partially redacted (`k***@gmail.com`, not full email).
- **Expiration:** Sellers should set `expires_at`. Agents should treat expired contexts as failed.

---

## ✅ Pre-Submission Checklist

- [x] I have created a GitHub Issue with the `SEP` and `proposal` tags (#142)
- [x] This PR includes a changelog entry file in `changelog/unreleased/intervention-types-extension.md`

---

## 📚 Additional Context

This SEP was motivated by the discussion in #142 and the arrival of SEP #141 (Stripe Link). The three types map to distinct agent capabilities, which is what capability negotiation (#137) actually cares about. An agent that handles redirects isn't necessarily capable of rendering a webview, and that distinction needs to be declarable at discovery time.

The `InterventionContext` object is related to the gap in #98 about how `intervention_required` errors surface in responses -- this SEP provides the structured payload that #98 identified as missing.

---

## 🙋 Questions for Reviewers

1. Should `webview` carry additional metadata (SDK URL, required dimensions, sandbox flags)?
2. Is `redirect` underspecified for the popup/modal vs full-navigation distinction? Should there be sub-types or should `display_context` handle that?
3. Should `InterventionResponse` be on the complete request too, or just update?
4. Does the Stripe Link team (cc @prasad-stripe) see `otp` as sufficient for the Link flow, or does Link need additional fields?

### DIFF
--- a/changelog/unreleased/intervention-types-extension.md
+++ b/changelog/unreleased/intervention-types-extension.md
@@ -1,0 +1,8 @@
+### Added
+- New intervention types: `redirect`, `webview`, `otp` in `InterventionCapabilities.supported` and `InterventionCapabilities.required` enums
+- `InterventionContext` schema for structured intervention metadata on checkout session responses
+- `InterventionResponse` schema for agent submission of intervention results (OTP codes, redirect completion)
+- `intervention_context` field on `CheckoutSessionBase` for pending intervention details
+- `intervention_response` field on `CheckoutSessionUpdateRequest` for agent intervention submissions
+- Examples for OTP flow, redirect flow, and capability mismatch scenarios
+- RFC: `rfcs/rfc.intervention_types.md`
--- a/examples/unreleased/intervention-types/examples.intervention_types.json
+++ b/examples/unreleased/intervention-types/examples.intervention_types.json
@@ -1,0 +1,389 @@
+{
+  "otp_flow_create_request": {
+    "currency": "usd",
+    "line_items": [
+      {
+        "id": "item_sneakers_001"
+      }
+    ],
+    "capabilities": {
+      "interventions": {
+        "supported": ["3ds", "otp"],
+        "display_context": "native",
+        "redirect_context": "none",
+        "max_redirects": 0,
+        "max_interaction_depth": 2
+      }
+    },
+    "fulfillment_details": {
+      "name": "Jane Smith",
+      "phone_number": "15559876543",
+      "email": "jane@example.com",
+      "address": {
+        "name": "Jane Smith",
+        "line_one": "456 Commerce Ave",
+        "line_two": "Apt 12",
+        "city": "Austin",
+        "state": "TX",
+        "country": "US",
+        "postal_code": "78701"
+      }
+    }
+  },
+
+  "otp_flow_session_response_intervention_required": {
+    "id": "checkout_session_otp_001",
+    "protocol": {
+      "version": "2026-01-30"
+    },
+    "status": "authentication_required",
+    "currency": "usd",
+    "line_items": [
+      {
+        "id": "item_sneakers_001",
+        "name": "Running Sneakers",
+        "description": "Lightweight running shoes, size 10",
+        "quantity": "1",
+        "unit_price": 12999,
+        "images": [
+          {
+            "url": "https://shop.example.com/images/sneakers-001.jpg",
+            "alt_text": "Running sneakers"
+          }
+        ]
+      }
+    ],
+    "totals": [
+      {
+        "type": "subtotal",
+        "label": "Subtotal",
+        "amount": 12999
+      },
+      {
+        "type": "shipping",
+        "label": "Standard Shipping",
+        "amount": 599
+      },
+      {
+        "type": "tax",
+        "label": "Sales Tax",
+        "amount": 1072
+      },
+      {
+        "type": "total",
+        "label": "Total",
+        "amount": 14670
+      }
+    ],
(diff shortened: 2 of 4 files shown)


sioked:
This PR does a nice job codifying the intervention vocabulary and I like that this can be used to add redirect/webview interventions to the interactions. One architectural question worth discussing before this lands:

The PR places `intervention_context` on `CheckoutSessionBase` and `intervention_response` on `CheckoutSessionUpdateRequest`, coupling the intervention lifecycle to the core checkout session. But the motivation section calls out wallet handlers specifically — Stripe Link, Google Pay, Shop Pay — and for those, the intervention creates a sequencing issue.

Take the Stripe Link OTP flow as described in the PR: the agent calls update with an `intervention_response` containing the OTP code, the merchant resolves it, and the session moves to `ready_for_payment`. But at that point the buyer hasn't selected a payment method yet — they complete OTP authentication and are then presented with their saved payment options by the **payment handler**, not the checkout session. The handler presents options, the buyer picks one, a token is issued, and only then can the agent call `complete`. So the checkout session is managing intervention state for a flow whose next step is owned by the payment handler. The intervention context passes through the merchant, but the merchant isn't the one driving what happens after it resolves.

3DS is different — it fires after the agent submits a token in `complete`, so it naturally belongs in the checkout session's error/retry cycle. The wallet auth flows the PR is targeting are different: they are pre-tokenization flows owned by the payment handler, and routing their context through the checkout session creates an awkward coupling in ownership to the checkout session.

Am I reading this correctly or am I missing something obvious in the PR?

Issue #187, opened 2026-03-20
SEP: Cart Capability: Pre-Checkout Basket Building


nvp-stripe:
# SEP: Cart Capability for Pre-Checkout Basket Building

## SEP Metadata

- **SEP Number**: #187
- **Status**: `proposal` (awaiting sponsor)
- **Type**: [x] Major Change [ ] Process Change
- **Related Issues/RFCs**: `rfcs/rfc.cart.md`

---

## Abstract

This SEP introduces a **Cart** capability to ACP -- a lightweight CRUD interface for pre-checkout basket building. Agents frequently build baskets incrementally ("add this, remove that, what's the total?") before the buyer expresses purchase intent. Today, ACP requires creating a full checkout session for this, which triggers capability negotiation, payment handler configuration, and status lifecycle management -- all unnecessary overhead for browsing.

The cart provides four endpoints (`POST /carts`, `GET /carts/{id}`, `POST /carts/{id}`, `POST /carts/{id}/cancel`) with estimated pricing, session persistence, and a `cart_id` field on checkout create for seamless cart-to-checkout conversion. No new domain types are introduced; carts reuse existing ACP types (LineItem, Buyer, Total, Message).

---

## Motivation

ACP's checkout session is designed for purchase finalization, but agents need a pre-purchase exploration phase. Without a cart:

- **Agents must create heavyweight checkout sessions for browsing** -- triggering capability negotiation, payment handler setup, and status lifecycle for what is essentially adding items to a basket.
- **Agents must track items client-side** -- meaning the seller has no visibility into browsing behavior, cannot provide estimated pricing, and cannot validate item availability until checkout.
- **No session persistence** -- there's no way to save a basket across agent sessions or share it via a URL for human-in-the-loop flows.
- **No incremental building** -- the common pattern of "add this... add that... remove that... check out" doesn't map cleanly to checkout session semantics.

Carts solve these by providing a binary-state (exists / not found) resource with estimated totals, no payment configuration, and no status lifecycle. The typical flow becomes: cart (browsing) -> checkout session (purchasing) -> order.

---

## Specification

This PR includes the following spec changes:

**New files:**
- `rfcs/rfc.cart.md` -- Full RFC with design, operations, schema, and conformance checklist
- `spec/unreleased/json-schema/schema.cart.json` -- Cart, CartCreateRequest, CartUpdateRequest schemas (all referencing existing ACP types via `$ref`)
[...]

PR #188, opened 2026-03-20
SEP: Add Cart Capability

nvp-stripe:
# SEP: Cart Capability for Pre-Checkout Basket Building

## 📋 SEP Metadata

- **SEP Number**: #187
- **Status**: `proposal`
- **Related Issues/RFCs**: `rfcs/rfc.cart.md`

---

## 🎯 Abstract

This SEP introduces a **Cart** capability to ACP -- a lightweight CRUD interface for pre-checkout basket building. Agents frequently build baskets incrementally ("add this, remove that, what's the total?") before the buyer expresses purchase intent. Today, ACP requires creating a full checkout session for this, which triggers capability negotiation, payment handler configuration, and status lifecycle management -- all unnecessary overhead for browsing.

The cart provides four endpoints (`POST /carts`, `GET /carts/{id}`, `POST /carts/{id}`, `POST /carts/{id}/cancel`) with estimated pricing and session persistence. No new domain types are introduced; carts reuse existing ACP types (LineItem, Buyer, Total, Message).

---

## 💡 Motivation

ACP's checkout session is designed for purchase finalization, but agents need a pre-purchase exploration phase. Without a cart:

* **Agents must create heavyweight checkout sessions for browsing** -- triggering capability negotiation, payment handler setup, and status lifecycle for what is essentially adding items to a basket.
* **Agents must track items client-side** -- meaning the seller has no visibility into browsing behavior, cannot provide estimated pricing, and cannot validate item availability until checkout.
* **No session persistence** -- there's no way to save a basket across agent sessions or share it via a URL for human-in-the-loop flows.
* **No incremental building** -- the common pattern of "add this... add that... remove that... check out" doesn't map cleanly to checkout session semantics.

Carts solve these by providing a binary-state (exists / not found) resource with estimated totals, no payment configuration, and no status lifecycle. The typical flow becomes: cart (browsing) -> checkout session (purchasing) -> order.

---

## 📐 Specification

This PR includes the following spec changes:

**New files:**

* `rfcs/rfc.cart.md` -- Full RFC with design, operations, schema, and conformance checklist
* `spec/unreleased/json-schema/schema.cart.json` -- Cart, CartCreateRequest, CartUpdateRequest schemas (all referencing existing ACP types via `$ref`)
* `spec/unreleased/openapi/openapi.cart.yaml` -- OpenAPI 3.1 for 4 REST endpoints
* `changelog/unreleased/cart.md` -- Changelog entry

**Modified files:**

* `spec/unreleased/json-schema/schema.agentic_checkout.json` -- Added `"carts"` to discovery services enum
* `rfcs/rfc.discovery.md` -- Added `"carts"` to services enum table and example

**Key design decisions:**

* **4 operations**: Create (`POST /carts`), Get (`GET /carts/{id}`), Update (`POST /carts/{id}`), Cancel (`POST /carts/{id}/cancel`)
* **Full replacement on update** -- matches existing ACP checkout update semantics; documented rationale and future ETag/`If-Match` option for optimistic concurrency
* **Single-seller scoping** -- carts are scoped to one seller; agents maintain separate carts per merchant for cross-merchant shopping
* **External cart origins** -- carts MAY originate outside the ACP API (e.g., merchant storefronts); agents can pick up browser-created carts via `GET /carts/{id}`
* **Manual cart-to-checkout conversion** -- agent reads cart via `GET /carts/{id}` and creates a checkout session with the retrieved items
* **No capabilities, no payment, no status lifecycle** -- these are checkout concerns

**Deferred to follow-up SEPs:**

* Cart-to-checkout conversion via a `cart_id` field on `CheckoutSessionCreateRequest` (conversion rules, idempotency, post-conversion cart lifecycle)
* Extension support (e.g., discount codes during browsing, fulfillment previews)

---

## 🤔 Rationale

**Why a separate resource instead of a "lightweight checkout"?**
Overloading checkout with a "pre-payment" mode would complicate the status lifecycle and capability negotiation contract. A separate cart resource with no status machine is simpler for both implementers and agents.

**Why full replacement instead of partial updates (add/remove item)?**
Consistent with ACP's existing checkout update pattern. Full replacement is simpler to implement and avoids the complexity of item-level CRUD operations (add, remove, change quantity as separate calls). The agent always has the full desired state and sends it. Future protocol versions MAY introduce optimistic concurrency (e.g., `If-Match` / ETag) if race conditions prove problematic in practice.

**Why reuse existing types instead of cart-specific types?**
Carts and checkouts share the same commerce domain (items, buyers, totals). Reusing types via `$ref` ensures consistency and reduces the schema surface area.

**Why is `currency` not required on cart create (unlike checkout)?**
Carts are lightweight. The seller determines currency from context (buyer locale, buyer address, storefront default, or seller default). If the seller cannot determine a currency, it SHOULD return `422 Unprocessable Entity`. Currency is always present on the cart response.

**Why defer cart-to-checkout conversion and extensions?**
Keeping the initial SEP focused on the core CRUD resource reduces scope and implementation complexity. The manual conversion path (read cart, create checkout with items) works today. `cart_id` conversion and extension support can be added additively in follow-up SEPs without breaking changes.

---

## 🔄 Backward Compatibility

This is a purely additive change with no breaking changes:

* New endpoints (`/carts/*`) do not conflict with existing paths.
* `"carts"` in the discovery services enum is a new value; existing values are unaffected.
* Cart is an optional service; sellers that don't implement it simply omit `"carts"` from their discovery document.
* No fields are added to or modified on existing ACP types. The checkout session schema is unchanged.

No deprecation or migration is needed.

---

Not yet available. The RFC and schema definitions in this PR serve as the specification from which implementations can be built.

---

## 🔒 Security Implications

* **Authentication**: Cart endpoints use the same `Authorization: Bearer` mechanism as checkout. No new auth flows are introduced.
* **Data privacy**: Carts carry buyer email and item data, same sensitivity as checkout sessions. No additional PII exposure.
* **Cart enumeration**: Cart IDs are server-generated opaque strings. Sequential or predictable IDs SHOULD be avoided to prevent enumeration.
* **Expiry**: Carts support `expires_at` to prevent indefinite data retention. Sellers SHOULD set reasonable TTLs and SHOULD refresh expiry on successful updates.

---

## ✅ Pre-Submission Checklist

- [x] This PR includes example requests/responses (in the RFC and OpenAPI)
- [x] This PR includes a changelog entry file in `changelog/unreleased/cart.md`

---

## 📚 Additional Context

The cart capability was designed based on common agent interaction patterns observed in early ACP implementations, where agents frequently needed to build baskets incrementally before committing to a purchase. The design prioritizes simplicity and reuse of existing ACP types to minimize the implementation surface area for sellers.

The scope was intentionally narrowed to focus on the core cart CRUD resource. Cart-to-checkout conversion (`cart_id` on checkout create) and extension support (e.g., discounts on carts) are deferred to follow-up SEPs that will build on this foundation additively.

---

## 🙋 Questions for Reviewers

1. **Update semantics**: Should cart update be `POST` (matching checkout) or `PUT` (more RESTful for full replacement)? The current proposal uses `POST` for consistency with ACP's existing patterns.
2. **Cart expiry defaults**: Should the spec recommend a default TTL (e.g., 24 hours, 7 days), or leave this entirely to the seller?
3. **Conversion path**: Is the manual cart-to-checkout conversion (read cart, create checkout) sufficient for the initial release, or should `cart_id` conversion be included in this SEP?

### DIFF
--- a/docs/sep-cart-capability.md
+++ b/docs/sep-cart-capability.md
@@ -1,0 +1,149 @@
+# SEP: Cart Capability for Pre-Checkout Basket Building
+
+## SEP Metadata
+
+- **SEP Number**: #187
+- **Status**: `proposal`
+- **Type**: [x] Major Change [ ] Process Change
+- **Related Issues/RFCs**: `rfcs/rfc.cart.md`
+
+---
+
+## Abstract
+
+This SEP introduces a **Cart** capability to ACP -- a lightweight CRUD interface for pre-checkout basket building. Agents frequently build baskets incrementally ("add this, remove that, what's the total?") before the buyer expresses purchase intent. Today, ACP requires creating a full checkout session for this, which triggers capability negotiation, payment handler configuration, and status lifecycle management -- all unnecessary overhead for browsing.
+
+The cart provides four endpoints (`POST /carts`, `GET /carts/{id}`, `PUT /carts/{id}`, `POST /carts/{id}/cancel`) with estimated pricing and session persistence. No new domain types are introduced; carts reuse existing ACP types (LineItem, Buyer, Total, Message).
+
+---
+
+## Motivation
+
+ACP's checkout session is designed for purchase finalization, but agents need a pre-purchase exploration phase. Without a cart:
+
+* **Agents must create heavyweight checkout sessions for browsing** -- triggering capability negotiation, payment handler setup, and status lifecycle for what is essentially adding items to a basket.
+* **Agents must track items client-side** -- meaning the seller has no visibility into browsing behavior, cannot provide estimated pricing, and cannot validate item availability until checkout.
+* **No session persistence** -- there's no way to save a basket across agent sessions or share it via a URL for human-in-the-loop flows.
+* **No incremental building** -- the common pattern of "add this... add that... remove that... check out" doesn't map cleanly to checkout session semantics.
+
+Carts solve these by providing a binary-state (exists / not found) resource with estimated totals, no payment configuration, and no status lifecycle. The typical flow becomes: cart (browsing) -> checkout session (purchasing) -> order.
+
+---
+
+## Specification
+
+This PR includes the following spec changes:
+
+**New files:**
+
+* `rfcs/rfc.cart.md` -- Full RFC with design, operations, schema, and conformance checklist
+* `spec/unreleased/json-schema/schema.cart.json` -- Cart, CartCreateRequest, CartUpdateRequest schemas (all referencing existing ACP types via `$ref`)
+* `spec/unreleased/openapi/openapi.cart.yaml` -- OpenAPI 3.1 for 4 REST endpoints
+* `changelog/unreleased/cart.md` -- Changelog entry
+
+**Modified files:**
+
+* `spec/unreleased/json-schema/schema.agentic_checkout.json` -- Added `"carts"` to discovery services enum
+* `rfcs/rfc.discovery.md` -- Added `"carts"` to services enum table and example
+
+**Key design decisions:**
+
+* **4 operations**: Create (`POST /carts`), Get (`GET /carts/{id}`), Update (`PUT /carts/{id}`), Cancel (`POST /carts/{id}/cancel`)
+* **Full replacement on update** -- matches existing ACP checkout update semantics
+* **Estimated (non-authoritative) pricing** -- totals on a cart are estimates; authoritative pricing comes at checkout
+* **Session persistence** -- carts survive across agent sessions and can be shared via `continue_url`
+* **No capabilities, no payment, no status lifecycle** -- these are checkout concerns
+* **Binary state** -- a cart either exists or returns 404; there is no status machine
+
+**Cart response object fields:**
+
+* `id` (required) -- server-generated unique identifier
+* `line_items` (required) -- reuses existing ACP LineItem type
+* `buyer` (optional) -- reuses existing ACP Buyer type
+* `currency` (required) -- ISO 4217 currency code, determined by seller
+* `totals` (required) -- estimated cost breakdown (subtotal, tax if calculable, total)
+* `messages` (optional) -- MessageInfo, MessageWarning, MessageError
+* `continue_url` (optional) -- URI for cart handoff/session recovery
+* `expires_at` (optional) -- RFC 3339 expiry timestamp
+
+**Deferred to follow-up SEPs:**
+
+* **Cart-to-checkout conversion** via a `cart_id` field on `CheckoutSessionCreateRequest` -- agents currently convert carts to checkouts manually by reading the cart and creating a checkout session with the retrieved items
+* **Extension support** (e.g., discount codes during browsing) -- agents can apply discounts at checkout time instead
+
+This SEP establishes the core cart resource; conversion and extension semantics will build on it additively.
+
+---
+
+## Rationale
+
+**Why a separate resource instead of a "lightweight checkout"?**
+Overloading checkout with a "pre-payment" mode would complicate the status lifecycle and capability negotiation contract. A separate cart resource with no status machine is simpler for both implementers and agents.
+
+**Why full replacement instead of partial updates (add/remove item)?**
+Consistent with ACP's existing checkout update pattern. Full replacement is simpler to implement and avoids the complexity of item-level CRUD operations (add, remove, change quantity as separate calls). The agent always has the full desired state and sends it.
+
+**Why reuse existing types instead of cart-specific types?**
+Carts and checkouts share the same commerce domain (items, buyers, totals). Reusing types via `$ref` ensures consistency and reduces the schema surface area.
+
+**Why is `currency` not required on cart create (unlike checkout)?**
+Carts are lightweight. The seller determines currency from context (locale, buyer location) or defaults. Requiring currency upfront adds friction for browsing. Currency is always present on the response (determined by the seller).
+
+**Why defer cart-to-checkout conversion?**
+Keeping the initial SEP focused on the core CRUD resource reduces scope and implementation complexity. The manual conversion path (read cart, create checkout with items) works today. A `cart_id` field on checkout create can be added additively in a follow-up SEP without breaking changes.
+
+---
+
+## Backward Compatibility
+
+This is a purely additive change with no breaking changes:
+
+* New endpoints (`/carts/*`) do not conflict with existing paths.
+* `"carts"` in the discovery services enum is a new value; existing values are unaffected.
+* Cart is an optional service; sellers that don't implement it simply omit `"carts"` from their discovery document.
+
+No deprecation or migration is needed.
+
+---
+
+## Reference Implementation
+
+Not yet available. The RFC and schema definitions in this PR serve as the specification from which implementations can be built.
+
+---
+
+## Security Implications
+
+* **Authentication**: Cart endpoints use the same `Authorization: Bearer` mechanism as checkout. No new auth flows are introduced.
+* **Data privacy**: Carts carry buyer email and item data, same sensitivity as checkout sessions. No additional PII exposure.
+* **Cart enumeration**: Cart IDs are server-generated opaque strings. Sequential or predictable IDs SHOULD be avoided to prevent enumeration.
+* **Expiry**: Carts support `expires_at` to prevent indefinite data retention. Sellers SHOULD set reasonable TTLs.
+
+---
+
+## Pre-Submission Checklist
+
+- [x] I have created a GitHub Issue with the `SEP` and `proposal` tags
+- [x] I have linked the SEP issue number above
+- [x] I have discussed this proposal in the community (Discord/GitHub Discussions)
+- [x] I have signed the Contributor License Agreement (CLA)
+- [x] This PR includes updates to OpenAPI/JSON schemas
+- [x] This PR includes example requests/responses (in the RFC and OpenAPI)
+- [x] This PR includes a changelog entry file in `changelog/unreleased/cart.md`
+- [x] I am seeking or have found a sponsor (Founding Maintainer)
+
+---
+
+## Additional Context
+
+The cart capability was designed based on common agent interaction patterns observed in early ACP implementations, where agents frequently needed to build baskets incrementally before committing to a purchase. The design prioritizes simplicity and reuse of existing ACP types to minimize the implementation surface area for sellers.
+
+The scope was intentionally narrowed to focus on the core cart CRUD resource. Cart-to-checkout conversion (`cart_id` on checkout create) and extension support (e.g., discounts on carts) are deferred to follow-up SEPs that will build on this foundation additively.
+
+---
+
+## Questions for Reviewers
+
+1. **Update semantics**: Should cart update be `POST` (matching checkout) or `PUT` (more RESTful for full replacement)? The current proposal uses `POST` for consistency with ACP's existing patterns.
+2. **Cart expiry defaults**: Should the spec recommend a default TTL (e.g., 24 hours, 7 days), or leave this entirely to the seller?
+3. **Conversion path**: Is the manual cart-to-checkout conversion (read cart, create checkout) sufficient for the initial release, or should `cart_id` conversion be included in this SEP?
--- a/rfcs/rfc.cart.md
+++ b/rfcs/rfc.cart.md
@@ -263,6 +263,8 @@ Future extensions MAY declare cart support by extending the `Cart` type, followi
 
 A follow-up SEP will also define cart-to-checkout conversion via a `cart_id` field on `CheckoutSessionCreateRequest`, including conversion rules, idempotency semantics, and post-conversion cart lifecycle.
 
+Future SEPs may also define `continue_url` as an input on `CartCreateRequest`, enabling agents to reference existing merchant-created carts by URL without requiring knowledge of individual product IDs.
+
 ---
 
 ## 6. Discovery Integration
--- a/rfcs/rfc.cart.md
+++ b/rfcs/rfc.cart.md
@@ -1,7 +1,7 @@
 # RFC: Agentic Commerce — Cart
 
-**Status:** Proposal  
-**Version:** unreleased  
+**Status:** Proposal
+**Version:** unreleased
 **Scope:** Pre-checkout basket building for the Agentic Commerce Protocol
 
 This RFC introduces a **Cart** capability for the Agentic Commerce Protocol (ACP). Carts provide a lightweight CRUD interface for item collection before purchase intent is established, enabling incremental basket building, estimated pricing, and session persistence without the overhead of a full checkout lifecycle.
@@ -44,10 +44,10 @@ Neither approach serves agents or sellers well.
 
 1. **Simple CRUD**: Create, read, update, and cancel carts with minimal required fields.
 2. **Reuse existing types**: Cart line items, buyer, totals, and messages reuse existing ACP types — no new domain concepts.
-3. **Cart-to-checkout conversion**: A single `cart_id` field on checkout create converts a cart into a checkout session, carrying over items, buyer info, and context.
-4. **Estimated totals**: Sellers provide best-effort pricing estimates. Totals are explicitly non-authoritative until checkout.
-5. **Optional capability**: Cart is an optional ACP service. Sellers MAY implement it. Agents MUST NOT assume cart support.
-6. **Extension support**: Carts support the discount extension, enabling agents to apply discount codes during browsing.
+3. **Estimated totals**: Sellers provide best-effort pricing estimates. Totals are explicitly non-authoritative until checkout.
+4. **Optional capability**: Cart is an optional ACP service. Sellers MAY implement it. Agents MUST NOT assume cart support.
+
+> **Deferred to follow-up SEPs:** Cart-to-checkout conversion via a `cart_id` field on `CheckoutSessionCreateRequest`, and extension support (e.g., discount codes during browsing) are intentionally deferred. This SEP establishes the core cart resource; conversion and extension semantics will build on it additively.
 
 ### 2.2 Non-Goals
 
@@ -55,11 +55,14 @@ Neither approach serves agents or sellers well.
 - **Status lifecycle**: Carts do not have a status state machine. A cart either exists or it does not.
 - **Capability negotiation**: Carts do not carry a `capabilities` object. Capability negotiation occurs at checkout creation time.
 - **Fulfillment option selection**: Carts do not support fulfillment option selection. Fulfillment options are presented and selected during checkout.
-- **Order creation**: Carts cannot be completed or converted to orders directly. They must first be converted to a checkout session.
+- **Order creation**: Carts cannot be completed or converted to orders directly.
+- **Cart-to-checkout bridge**: Automatic conversion of carts to checkout sessions via a `cart_id` field is deferred to a follow-up SEP. Agents can manually create checkout sessions using cart contents retrieved via `GET /carts/{id}`.
 
 ---
 
 ## 3. Design
+
+**Scope:** Carts are scoped to a single seller. Each cart is hosted under the seller's `api_base_url`. Agents managing cross-merchant shopping SHOULD maintain separate carts per seller.
 
 ### 3.1 Cart vs Checkout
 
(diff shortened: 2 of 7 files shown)


lhwa:
# Review Comments: Cart Capability SEP (#187 / PR #188)

**SEP Issue:** https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/187
**SEP PR:** https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188
**Author:** nvp-stripe
**Reviewer:** Lee Hwa (lhwa)

---

## Overall

Support this proposal. Separating cart from checkout session is the right architectural decision. The comments below are for clarification, completion, and forward-looking considerations.

---

## Comment 1: Cart Lifecycle After Checkout Conversion

When `cart_id` is passed to `CheckoutSessionCreateRequest`, the spec says conversion is idempotent. But the cart's post-conversion state is undefined.

**Questions:**
- Is the cart frozen after conversion? Can it still be retrieved?
- Can a second checkout be created from the same cart (e.g., after abandonment)?
- Does updating the cart after conversion affect the in-progress checkout?

**Suggestion:** Define post-conversion cart state explicitly — e.g., "Once converted, the cart is frozen. Updates return 409 Conflict. The cart remains retrievable for reference."

---

## Comment 2: Browser-Created Carts and Agent-Driven Checkout

The SEP assumes carts are created by agents via the API. A high-value missing scenario: **carts created by users in a browser, then handed off to an agent for checkout.**

**Why this matters:**
- **No catalog sync needed**: The merchant's storefront handles product discovery; the agent doesn't need a replicated catalog.
- **Full merchant features preserved**: Personalization, custom configurators, loyalty pricing, and bundle deals work natively — these are extremely difficult to replicate via API.
- **Smaller agent surface area**: Agents focus on checkout and payment, not storefront replication.
- **Incremental adoption**: Merchants with existing storefronts can adopt ACP without rebuilding their browsing experience.

**What the spec needs to address:**
- **Cart ID portability**: Can a merchant-generated cart ID be used in an ACP `CheckoutSessionCreateRequest`?
- **Auth handoff**: How does the agent access a cart created in a different session?
- **Cart visibility**: Can the agent `GET /carts/{id}` on a browser-created cart before converting to checkout?

**Suggestion:** Add language acknowledging that carts MAY originate outside the ACP API and that `cart_id` on checkout SHOULD support externally-created carts. This positions ACP as complementary to existing commerce, not a full replacement.

---

## Comment 3: Single-Merchant vs. Multi-Merchant Carts

The SEP doesn't state whether a cart is scoped to a single merchant or can span multiple merchants.

**Single-merchant** (implicit in the current design — endpoints live under each seller's `api_base_url`) is simpler, with 1:1 cart-to-checkout mapping.

**Multi-merchant** reflects how agents actually shop: building a unified cart across merchants, then fanning out to separate checkouts at payment time. This is a core agentic commerce value prop — abstracting away multi-merchant complexity for the buyer.

**Questions:**
- Is single-merchant scoping intentional? If so, state it explicitly and document how agents should manage cross-merchant scenarios.
- Does the schema support future extension to multi-merchant? Should `line_items` include a `seller_id`?
- How does multi-merchant interact with cart-to-checkout conversion (fan-out to multiple sessions)?

**Suggestion:** At minimum, clarify: "Carts are scoped to a single seller. Agents managing cross-merchant shopping SHOULD maintain separate carts per seller." Note multi-merchant as a future extension point if on the roadmap.

---

## Comment 4: Full-Replacement Semantics on Update

`CartUpdateRequest` uses full replacement of `line_items`, requiring the agent to track full cart state client-side.

**Concerns:**
- **Race conditions**: Concurrent updates from multi-turn or multi-agent flows can silently overwrite each other.
- **Payload size**: Sending the full item list on every single-item change is inefficient for large carts.

**Note:** PR #188 explains the rationale: full replacement is simpler and consistent with ACP's checkout update pattern. However, carts are updated far more frequently than checkouts, making these trade-offs more pronounced. Consider documenting this design rationale explicitly in the RFC.

---

## Comment 5: Multi-Currency Resolution

`currency` is optional on `CartCreateRequest` but required on the `Cart` response. The spec doesn't define how the seller resolves currency when omitted.

**Suggestion:** Document the resolution order — e.g., buyer locale/address → seller default → 422 error.

---

## Comment 6: Discount Carryover to Checkout

Supporting discounts at cart stage is great — agents can validate promo codes before checkout. But the conversion behavior is unspecified.

**Questions:**
- Are discounts carried over automatically when converting via `cart_id`?
- Must the agent re-apply discount codes at checkout?
- What happens if a discount expires between cart creation and checkout conversion?

---

## Comment 7: Cart Expiry and Session Management

`expires_at` is returned but there's no mechanism to extend a cart's lifetime. Real shopping flows can span hours or days as users pause, compare, or context-switch.

**Questions:**
- Should a no-op update (e.g., `POST /carts/{id}` with unchanged items) refresh expiry?
- Or is expiry-and-recreate the intended pattern? If so, should agents persist cart contents client-side?

---

## Comment 8: Quantity Updates Without Full Item Replacement

Quantity changes are the most frequent cart mutation in commerce. The current design requires reconstructing the entire `line_items` array to change a single item's quantity. Related to Comment 4, but worth highlighting — the simplest operation requires the most work from the agent.

---

## Summary

| # | Topic | Type | Priority |
|---|-------|------|----------|
| 1 | Post-conversion cart lifecycle | Spec gap | High |
| 2 | Browser-created cart → agent checkout | Missing scenario | High |
| 3 | Single-merchant vs. multi-merchant carts | Design question | High |
| 4 | Full-replacement vs. partial update semantics | Design question | Medium |
| 5 | Currency resolution when omitted | Spec gap | Medium |
| 6 | Discount carryover to checkout | Spec gap | Medium |
| 7 | Cart expiry extension mechanism | Gap | Low |
| 8 | Quantity update ergonomics | Design question | Low |

---

*Prepared for posting as review comments on PR #188. Comments can be posted individually or as a single review.*

nvp-stripe:
Thanks for the thorough review @lhwa. Really appreciate the detail here! A couple of these hit on features that were deferred after the version you reviewed, so let me address each one.

---

### Comment 1: Cart Lifecycle After Checkout Conversion

This is deferred. The current SEP does **not** include `cart_id` on `CheckoutSessionCreateRequest` — that's explicitly called out as a follow-up SEP (see [RFC §2.2 Non-Goals](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md) and [§5 Future Extensions](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md)).

The conversion path today is manual: agent calls `GET /carts/{id}`, reads the items, and creates a checkout session with those items. No cart state change occurs — the cart remains retrievable and updatable.

Your questions about post-conversion state (frozen? re-convertible? update-after-conversion?) are exactly the right ones — they'll need to be answered in the `cart_id` conversion SEP. I'll reference your comment there. For now, the cart is just a bag of items with no awareness of checkout.

---

### Comment 2: Browser-Created Carts and Agent-Driven Checkout

Great point, and the current design already supports this via `continue_url`. A merchant's storefront can create a cart through its own backend (using the same `/carts` API or its own internal system), then surface the `continue_url` or cart ID to the agent for retrieval and checkout.

Specifically:
- **Cart ID portability**: Yes — `GET /carts/{id}` works regardless of who created the cart. The cart ID is an opaque server-generated string; nothing in the spec restricts creation to API callers.
- **Auth handoff**: The agent authenticates with a Bearer token like any other ACP call. The seller's auth layer determines whether the agent has access to a given cart — this is the same model as checkout sessions.
- **Cart visibility**: Yes, `GET /carts/{id}` returns the full cart state to any authorized caller.

That said, I think you're right that we should make this explicit. I'll add language to the RFC acknowledging that carts MAY originate outside the ACP API (e.g., merchant storefronts) and that `GET /carts/{id}` SHOULD work for externally-created carts. Good call.

---

### Comment 3: Single-Merchant vs. Multi-Merchant Carts

Single-merchant is intentional and already stated in the RFC. From [§3 Design](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md):

> **Scope:** Carts are scoped to a single seller. Each cart is hosted under the seller's `api_base_url`. Agents managing cross-merchant shopping SHOULD maintain separate carts per seller.

Multi-merchant carts would require a fundamentally different architecture (aggregation layer, split checkout, cross-seller settlement) that's well beyond cart CRUD. Agree this could be a future extension point, but it's a separate concern from the cart resource itself.

---

### Comment 4: Full-Replacement Semantics on Update

Acknowledged — this is a deliberate trade-off documented in [§4.4](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md):

> Full replacement is intentional and consistent with ACP's checkout update pattern. While partial update operations (add item, remove item, change quantity) would reduce per-request payload, they significantly increase the API surface area and introduce new failure modes (concurrent add/remove, quantity delta ambiguity, item-level addressing). Agents maintain client-side cart state from the previous response and can trivially reconstruct the full items array. Future protocol versions MAY introduce optimistic concurrency (e.g., `If-Match` / ETag) if race conditions prove problematic in practice.

Your concern about race conditions is valid for multi-agent flows. The RFC already notes ETag/`If-Match` as a future option. For V1, the simplicity win of full replacement outweighs the concurrency risk — most cart interactions are single-agent, single-session.

---

### Comment 5: Multi-Currency Resolution

Already addressed in the RFC. From [§4.2 Create Cart](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md):

> **Currency resolution:** The `currency` field is not required on the create request. The seller determines currency from available context (buyer locale, buyer address, storefront default, or seller default). If the seller cannot determine a currency, it SHOULD return `422 Unprocessable Entity` with a message requesting the agent provide a `locale` or `buyer` with sufficient context.

This gives sellers flexibility while providing a clear fallback (422) when context is insufficient.

---

### Comment 6: Discount Carryover to Checkout

Deferred — same as Comment 1. Extension support (including discounts) is explicitly out of scope for this SEP. From the RFC:

> Extension integration with carts (e.g., discount codes during browsing, fulfillment previews) is deferred to follow-up SEPs.

Your questions about carryover semantics, re-application, and expiry are all things the extension follow-up will need to address.

---

### Comment 7: Cart Expiry and Session Management

The RFC addresses this in [§4.4](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/188/files#diff-rfcs/rfc.cart.md):

> **Cart expiry:** Sellers SHOULD refresh `expires_at` on any successful update. If a cart has expired, subsequent operations return `404 Not Found`; the agent MAY create a new cart with the previously-known items.

So yes — any update (including a no-change update) SHOULD refresh the expiry. And if a cart does expire, the agent recreates it. Since the agent always has the full cart state from the last response, recreation is trivial.

---

### Comment 8: Quantity Updates Without Full Item Replacement

This is the same trade-off as Comment 4. The agent has the full `line_items` array from the last `GET` or `POST` response. Changing a quantity means updating one element in that array and sending it back. It's a few extra bytes on the wire but dramatically simpler API surface.

If this becomes a real pain point in practice (very large carts, high-frequency updates), we can add partial update operations in a future version. For V1, full replacement keeps the spec small.

---

rfcs/rfc.cart.md
```diff
@@ -0,0 +1,337 @@
...
+  ],
+  "buyer": {
+    "email": "buyer@example.com"
+  },
+  "currency": "usd",
+  "totals": [
+    { "type": "subtotal", "amount": 25500 },
+    { "type": "total", "amount": 25500 }
+  ],
+  "continue_url": "https://seller.example.com/cart/cart_abc123",
```
lhwa, line 203: I see `continue_url` is added on the response — I assume this is intended for the seller to return a URL for sharing/handoff. But the highest-value handoff scenario is the reverse:
   a user builds a cart on the merchant's website and hands it to an agent for checkout.

  Suggest accepting `continue_url` as an input on `CartCreateRequest`, mutually exclusive with `line_items`. The flow would be:

  1. User browses merchant website, adds items to cart
  2. User tells agent: "check out my cart at merchant.com"
  3. Agent creates an ACP cart by passing the browser cart URL — no need to know individual product IDs:

      ```json
      POST /carts
      {
        "continue_url": "https://seller.example.com/cart/cart_abc123",
        "buyer": {
          "email": "buyer@example.com"
        }
      }
      ```

  4. Seller resolves the URL to the existing cart contents, returns a full `Cart` response with `line_items`, `totals`, etc.

  This is powerful for agents because:
  - **No seller catalog sync needed** — the agent doesn't need to know product IDs or SKUs. This addresses one of the biggest adoption barriers: agents today must either obtain the
  full merchant catalog or integrate with a product feed API before they can even create a cart.
  - **Preserves merchant-side state** — loyalty pricing, bundle deals, custom configurations all carry over. These are extremely difficult to replicate via API and would be lost if
  the agent had to reconstruct the cart item-by-item.
  - **No catalog data sync burden** — agents don't need to continuously poll for price changes, availability updates, or SKU mappings when the merchant's own cart URL already
  reflects the current state.

lhwa: This scenario does not block the current SEP.  We can defer this support to future extensions.

lhwa (commented):
@nvp-stripe Thank you for the comment by comment response. You have addressed my comments with very clear explanations and corresponding updates. 

I added an inline comment about `continue_url` — I think the current change on the response is not exactly the scenario I brought up initially (browser-created carts handed off to agents).  I would suggest `continue_url` be added as a cart creation input as well, so agents can take an existing merchant cart by URL without needing to know individual product IDs.

That said, this scenario does not need to block this SEP and we can defer to future extensions.

rfcs/rfc.cart.md
nvp-stripe: I agree. The `continue_url` as input idea is interesting but would add real complexity:
  - Mutually exclusive field validation (continue_url XOR line_items)
  - URL resolution semantics (what if the URL is invalid, expired, or from a different seller?)
  - It's essentially a cart import/link mechanism, which is a distinct feature

nvp-stripe:
> @nvp-stripe Thank you for the comment by comment response. You have addressed my comments with very clear explanations and corresponding updates.
> 
> I added an inline comment about `continue_url` — I think the current change on the response is not exactly the scenario I brought up initially (browser-created carts handed off to agents). I would suggest `continue_url` be added as a cart creation input as well, so agents can take an existing merchant cart by URL without needing to know individual product IDs.
> 
> That said, this scenario does not need to block this SEP and we can defer to future extensions.

Thanks for the thoughtful suggestion on continue_url as input. I agree this would be a valuable addition for the browser-to-agent handoff flow. I've added an explicit callout in the Future Extensions section (Section 5) acknowledging this as a candidate for a follow-up SEP. Keeping it out of scope here lets us ship the core cart CRUD cleanly, but it's now documented so it doesn't get lost.

rfcs/rfc.cart.md
```diff
@@ -0,0 +1,339 @@
...
+
+`POST /carts`
+
+**Request:**
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `line_items` | `Item[]` | Yes | Items to add to the cart. |
+| `buyer` | `Buyer` | No | Buyer information for personalized estimates. |
+| `locale` | string | No | Locale code for content localization. |
```
aravindrao-openai, line 141: Here and everywhere else, why are we introducing the concept of locale vs using Accept-Language in the header?

rfcs/rfc.cart.md
```diff
@@ -0,0 +1,339 @@
...
+  ],
+  "buyer": {
+    "email": "buyer@example.com"
+  },
+  "currency": "usd",
+  "totals": [
+    { "type": "subtotal", "amount": 25500 },
+    { "type": "total", "amount": 25500 }
+  ],
+  "continue_url": "https://seller.example.com/cart/cart_abc123",
```
aravindrao-openai, line 203: I have a slightly different question regarding `continue_url` - There is a need to do so when negotiating checkout, especially in the completion stage (e.g. for interventions such as CVV stepup). What purpose would it serve in the cart building stage?

nvp-stripe:
@aravindrao-openai - could you please take a look?

docs/sep-cart-capability.md
```diff
@@ -0,0 +1,149 @@
...
+This SEP establishes the core cart resource; conversion and extension semantics will build on it additively.
+
+---
+
+## Rationale
+
+**Why a separate resource instead of a "lightweight checkout"?**
+Overloading checkout with a "pre-payment" mode would complicate the status lifecycle and capability negotiation contract. A separate cart resource with no status machine is simpler for both implementers and agents.
+
+**Why full replacement instead of partial updates (add/remove item)?**
```
lhwa, line 83: A downside of the full replacement approach: agents would need to build and maintain a local cart to track items in order to support full replacement.

rfcs/rfc.cart.md
```diff
@@ -0,0 +1,339 @@
...
+  "expires_at": "2026-04-01T12:00:00Z"
+}
+```
+
+**Error handling:**
+
+When all requested items are unavailable, the seller MAY return an error instead of creating a cart:
```
lhwa, line 210: What about the partial out-of-stock case? If some items in line_items are out of stock, does this API return a success response with a message, or does it return an error?  Can we have an example to cover this scenario?

rfcs/rfc.cart.md
```diff
@@ -0,0 +1,339 @@
...
+Full replacement of the cart. The agent MUST send the complete desired cart state. The provided `line_items` replace the existing cart contents.
+
+**Request:**
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `line_items` | `Item[]` | Yes | Complete list of items (replaces existing). |
+| `buyer` | `Buyer` | No | Updated buyer information. |
+
+**Response:** `200 OK` with updated `Cart` object.
```
lhwa, line 287: What happens if an item in the update is out of stock? Should seller reject the entire update request or      
update the cart with warning messages?

spec/unreleased/openapi/openapi.cart.yaml
```diff
@@ -0,0 +1,327 @@
...
+            examples:
+              update_quantity:
+                summary: Change quantity and add item
+                value:
+                  line_items:
+                    - id: item_123
+                      quantity: 3
+                    - id: item_789
+                      quantity: 1
+      responses:
```
lhwa, line 172: And error 422

rfcs/rfc.cart.md
nvp-stripe: Good question. The reason I went with locale in the body rather than Accept-Language is that this field represents the buyer's locale, not the API caller's. The agent making the request might be operating in en-US but the buyer could be in ja-JP, and that matters for things like currency resolution and how prices get displayed. Since the locale is really a property of the cart itself (it gets persisted with the resource and used for downstream calculations like tax), putting it in the body felt more appropriate than overloading a header thats designed for HTTP content negotiation.

Also worth noting that the existing checkout session spec already uses locale as a body field, so this keeps things consistent across ACP.

rfcs/rfc.cart.md
nvp-stripe: Yeah so the use case is actually pretty different from checkout. At checkout continue_url is about payment interventions like CVV stepup and 3DS redirects. At the cart stage its more about handoff and recovery scenarios. For example, an agent builds a cart and then sends the URL to the buyer so they can review it in their browser before committing to purchase. Or if the agent session gets interrupted, the URL lets them pick up where they left off.

We also want to support the reverse flow eventually where a user starts building a cart on a merchant website and then hands it off to an agent. The continue_url on the response is what makes that kind of sharing possible. Section 5 of the RFC goes into more detail on these use cases.

docs/sep-cart-capability.md
nvp-stripe: Yeah thats a fair point and we did discuss this tradeoff in the earlier round of comments. The thinking is that for V1 the simplicity gains outweigh this downside. In practice the agent always has the full cart state from whatever the last response was (create, get, or update), so its not like they need to build it from scratch. And if they do lose state for whatever reason, a single GET recovers everything.

We noted ETag/If-Match as a potential future addition if concurrency becomes an issue in practice but for now we think full replacement keeps the spec surface area small.

rfcs/rfc.cart.md
nvp-stripe: Great catch, this is a gap we should close. The spec currently covers the case where all items are unavailable (422) but doesnt say what happens when only some items are out of stock.

I think the right behavior is: the seller should still create the cart with all the requested items and return a 200 with messages indicating which items have availability issues. That way the agent stays in control and can decide whether to remove those items or keep them in the cart (maybe the buyer wants to wait). This is consistent with the existing message pattern in the spec where sellers use messages for validation warnings like low stock or price changes.

I'll add a clarifying paragraph and an example response showing this. Good callout.

rfcs/rfc.cart.md
nvp-stripe: Same answer as above basically. For updates, if some items in the replacement are out of stock the seller should accept the update and return the cart with messages about the unavailble items. Rejecting the entire update because one item is out of stock would leave the cart in its old state which would be pretty confusing, especially since the agent might have been trying to remove other items or change quantities at the same time.

I'll add an example for the update path too.

spec/unreleased/openapi/openapi.cart.yaml
nvp-stripe: Good catch, thats a bug. The create endpoint has 422 but the update endpoint is missing it. Will fix.

prasad-stripe (approved):
Thanks @nvp-stripe for thoughtful SEP. Thanks @aravindrao-openai and @lhwa for healthy discussion, we made some great decisions to defer more complicated items to next versions.

lhwa (approved):
The proposal looks good to me. Thank you for addressing all of the comments. It was a great discussion.

aravindrao-openai (approved):

Issue #189, opened 2026-03-24
[SEP] Product Feeds


aravindrao-openai:
# SEP: Product Feeds

## SEP Metadata

- **Author(s)**: Aravind Rao, Rahul Bansal
- **Status**: `proposal`
- **Type**: [x] Major Change [ ] Process Change

---

## Abstract

This SEP introduces **product feeds** as a new ACP endpoints for publishing,
retrieving, and incrementally updating structured product catalogs outside the
checkout session lifecycle. The proposal adds a new authenticated resource,
`product_feeds`, with APIs to create a feed, fetch feed metadata, read the
feed's products, and upsert products by stable product identifier.

The feed model gives ACP a standard way to exchange catalog data that agents,
seller platforms, marketplaces, and commerce infrastructure can consume before
checkout begins. It complements ACP's existing checkout and order flows: checkout
remains the source of transactional truth, while product feeds provide pre-purchase
catalog distribution and synchronization.

The proposed schemas define a canonical `Product` object with nested
`Variant`, pricing, availability, media, category, seller, and marketplace
metadata. Feeds may optionally be scoped by `target_country`, enabling sellers
to expose regional assortments or localized availability without requiring a
separate protocol.

---

### Why add product feeds to ACP?

1. **ACP currently lacks a standard catalog transport**: ACP can complete a
   purchase once an agent already knows what the buyer wants, but it does not
   define how structured product data is exchanged before checkout.

2. **Checkout line items are too late in the flow**: The checkout APIs describe
   selected items, prices, totals, and fulfillment for a transaction in
   progress. They do not solve catalog ingestion, indexing, or pre-purchase
   retrieval across a merchant's assortment.

3. **Discovery intentionally excludes product catalog data**: The discovery SEP
   correctly keeps `/.well-known/acp.json` small, public, and stable. That
   leaves a gap for authenticated product distribution, which this SEP fills
   without turning discovery into a catalog endpoint.

4. **Merchants need both full sync and incremental sync**: Some integrations
   want a complete catalog snapshot; others want to update only the changed
   products. ACP should support both patterns explicitly rather than forcing
   every integrator into a proprietary bulk API.

5. **Regional catalog variants are common**: Product availability, compliance,
   pricing presentation, and assortment often vary by market. A feed-level
   `target_country` allows ACP to model market-specific feeds cleanly.

### Why existing ACP concepts are insufficient

[...]

PR #190, opened 2026-03-24
SEP: Add Product Feeds API to Agentic Commerce Protocol

rb-oai:
# SEP: Product Feeds

## 📋 SEP Metadata
- **SEP Number**: [SEP: Product Feeds #189](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/189)
- **Author(s)**: Aravind Rao, Rahul Bansal
- **Status**: `proposal`
- **Type**: [X] Major Change [ ] Process Change

---

## 🎯 Abstract

This PR proposes an unreleased Feed API surface for publishing feed metadata and managing product catalogs in ACP.

The new API surface includes:

- `POST /feeds` to create feed metadata
- `GET /feeds/{id}` to retrieve feed metadata
- `GET /feeds/{id}/products` to retrieve the current product set
- `PATCH /feeds/{id}/products` to partially upsert products by `Product.id`

This PR also adds the supporting schema bundle and example payloads needed to define a reusable catalog representation for ACP outside the checkout session lifecycle.

---

## 💡 Motivation

ACP currently supports transactional flows well, but it does not yet provide a standard way to exchange structured product catalog data before checkout begins.

This PR addresses that gap by introducing a feed-oriented model for catalog distribution and synchronization. In particular, it supports:

- feed metadata retrieval
- full product-set reads
- incremental product updates by stable `Product.id`
- a reusable product model with variants, pricing, availability, media, seller, marketplace, and category metadata

This change is intentionally scoped to the base feed API only. Promotions are deferred to the stacked follow-up PR so the core catalog surface can land independently.

---

## 📐 Specification

This proposal makes the following normative additions in unreleased ACP:

1. Add a new `FeedMetadata` model for server-managed feed metadata
2. Add a canonical `Product` / `Variant` catalog schema bundle in `schema.feed.json`
3. Add `POST /feeds` for feed creation
4. Add `GET /feeds/{id}` for feed metadata retrieval
5. Add `GET /feeds/{id}/products` for reading the full current product set
6. Add `PATCH /feeds/{id}/products` for partial upsert semantics keyed by `Product.id`

The schema bundle includes reusable definitions for:

- `Description`
- `Price`
- `Availability`
- `Barcode`
- `Media`
- `VariantOption`
- `Category`
- `Link`
- `Seller`
- `Condition`
- `Measure`
- `ReferenceMeasure`
- `UnitPrice`
- `Variant`
- `Product`
- `FeedMetadata`
- request/response envelopes for create, list, and upsert operations

This PR also documents the offline ingestion format:

- `metadata.json` uses the `FeedMetadata` shape
- `products.jsonl` contains one `Product` object per line
- file ingestion performs full replacement of the feed's product set
- partial updates are only supported through `PATCH /feeds/{id}/products`

---

## 🤔 Rationale

A feed resource is a better fit than overloading checkout or discovery for catalog exchange.

This structure gives ACP:

- a dedicated resource for catalog synchronization
- support for both full retrieval and incremental updates
- a stable product and variant representation that can be reused across implementations
- room to layer promotions and other catalog-adjacent features in follow-up changes

Separating the base feed model from promotions also keeps this PR reviewable and makes the stack easier to reason about.

---

## 🔄 Backward Compatibility

This is a backward-compatible, additive change in unreleased ACP.

Migration impact:

- no existing ACP endpoints are removed
- no existing ACP schemas are redefined in this PR
- clients that do not use feeds are unaffected

Implementations that adopt this API can begin using the new `/feeds` endpoints and schema bundle without changing existing checkout flows.

---

This PR adds the unreleased specification and related examples/docs in:

- `spec/unreleased/openapi/openapi.feed.yaml`
- `spec/unreleased/json-schema/schema.feed.json`
- `examples/unreleased/examples.feed.json`
- `changelog/unreleased/add-feed-api.md`

---

## 🔒 Security Implications

This PR introduces a new catalog-bearing API surface, so implementers should treat feed contents as potentially sensitive merchant data.

Key considerations:

- feed metadata and product catalogs may include pricing, assortment, and availability data
- product payloads may contain external URLs and rich content that consumers should treat as untrusted input
- incremental upsert behavior should validate payloads strictly before accepting product updates

This PR defines the surface and schemas; implementation-specific authentication, authorization, and abuse controls should be enforced by ACP servers.

---

## 📚 Additional Context

This PR is the base layer for the Product Feeds SEP and intentionally excludes promotions.

Promotions are deferred to the stacked follow-up branch/PR so the foundational feed resource and product model can be reviewed independently first.

---

## 🙋 Questions for Reviewers

- Are `/feeds` the right endpoint names for this first pass, or should this land as `/product_feeds` to match the SEP wording more closely?
- Is the scope split between base feeds and promotions the right review boundary?
- Is whole-product upsert by `Product.id` sufficiently clear for implementers?

---

## ✅ Pre-Submission Checklist

Before submitting this SEP PR, ensure:

### DIFF
--- a/changelog/unreleased/add-feed-api.md
+++ b/changelog/unreleased/add-feed-api.md
@@ -1,0 +1,21 @@
+# Feed API
+
+**Added** – an unreleased Feed API surface for feed metadata and product catalog management.
+
+## New API Surface
+
+- `POST /feeds` to create feed metadata
+- `GET /feeds/{id}` to retrieve feed metadata
+- `GET /feeds/{id}/products` to retrieve the current product set
+- `PATCH /feeds/{id}/products` to partially upsert products by `Product.id`
+
+## File Ingestion Format
+
+- `metadata.json` uses the `FeedMetadata` shape
+- `products.jsonl` contains one `Product` object per line
+- file ingestion performs full replacement of the feed's product set
+- partial updates are only supported through `PATCH /feeds/{id}/products`
+
+## Deferred
+
+- promotions are intentionally deferred to a stacked follow-up PR
--- a/changelog/unreleased/add-feed-api.md
+++ b/changelog/unreleased/add-feed-api.md
@@ -4,10 +4,14 @@
 
 ## New API Surface
 
-- `POST /feeds` to create feed metadata
-- `GET /feeds/{id}` to retrieve feed metadata
-- `GET /feeds/{id}/products` to retrieve the current product set
-- `PATCH /feeds/{id}/products` to partially upsert products by `Product.id`
+These endpoints are hosted by Agents and called by merchants. The Feed API is a
+push model: merchants push product catalog metadata and product records to
+Agents, rather than Agents pulling catalog data from merchant-hosted endpoints.
+
+- `POST /feeds` for merchants to create feed metadata on an Agent-hosted feed service
+- `GET /feeds/{id}` for merchants to retrieve feed metadata from the Agent
+- `GET /feeds/{id}/products` for merchants to retrieve the current Agent-hosted product set
+- `PATCH /feeds/{id}/products` for merchants to partially upsert products by `Product.id` into the Agent-hosted feed
 
 ## File Ingestion Format
 
--- a/examples/unreleased/examples.feed.json
+++ b/examples/unreleased/examples.feed.json
@@ -1,0 +1,107 @@
+{
+  "create_feed_request_minimal": {
+    "target_country": "US"
+  },
+  "create_feed_response": {
+    "id": "feed_8f3K2x",
+    "target_country": "US",
+    "updated_at": "2026-03-01T00:00:00Z"
+  },
+  "get_feed_response": {
+    "id": "feed_8f3K2x",
+    "target_country": "US",
+    "updated_at": "2026-03-01T00:00:00Z"
+  },
+  "get_products_response": {
+    "products": [
+      {
+        "id": "prod_classic_tee",
+        "title": "Classic Tee",
+        "variants": [
+          {
+            "id": "sku123-red-s",
+            "title": "Classic Tee - Red / Small"
+          }
+        ]
+      }
+    ]
+  },
+  "upsert_products_request_single_product": {
+    "products": [
+      {
+        "id": "prod_classic_tee",
+        "title": "Classic Tee",
+        "description": {
+          "plain": "A classic cotton tee with a soft feel and relaxed fit."
+        },
+        "url": "https://merchant.com/products/classic-tee",
+        "media": [
+          {
+            "type": "image",
+            "url": "https://cdn.merchant.com/products/classic-tee/main.jpg",
+            "alt_text": "Classic Tee front view"
+          }
+        ],
+        "variants": [
+          {
+            "id": "sku124-red-m",
+            "title": "Classic Tee - Red / Medium",
+            "description": {
+              "plain": "Classic cotton tee in red, size medium."
+            },
+            "url": "https://merchant.com/products/classic-tee?variant=sku124-red-m",
+            "barcodes": [
+              {
+                "type": "GTIN",
+                "value": "00012345600029"
+              }
+            ],
+            "price": {
+              "amount": 1999,
+              "currency": "USD"
+            },
+            "availability": {
+              "available": false,
+              "status": "out_of_stock"
+            },
+            "variant_options": [
+              {
+                "name": "Color",
+                "value": "Red"
+              },
+              {
+                "name": "Size",
+                "value": "Medium"
+              }
+            ],
+            "media": [
+              {
+                "type": "image",
+                "url": "https://cdn.merchant.com/products/classic-tee/red-medium-1.jpg",
+                "alt_text": "Classic Tee in red, size medium"
+              }
+            ],
+            "seller": {
+              "name": "Example Merchant"
+            },
+            "marketplace": {
+              "name": "Example Marketplace"
+            }
+          }
+        ]
+      }
+    ]
+  },
+  "feed_error_not_found": {
+    "type": "invalid_request",
+    "code": "feed_not_found",
+    "message": "Feed not found",
+    "param": "id"
+  },
+  "feed_error_invalid_request": {
+    "type": "invalid_request",
+    "code": "invalid_product_payload",
+    "message": "products must contain at least one valid Product",
+    "param": "products"
+  }
+}
--- a/spec/unreleased/json-schema/schema.feed.json
+++ b/spec/unreleased/json-schema/schema.feed.json
@@ -1,0 +1,626 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/schemas/feed/bundle.schema.json",
+  "title": "Feed - Schema Bundle",
+  "$defs": {
+    "Description": {
+      "description": "Structured long-form or rich-text description content for a product or variant.",
+      "type": "object",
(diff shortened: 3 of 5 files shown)


spec/unreleased/json-schema/schema.feed.json
```diff
@@ -0,0 +1,626 @@
...
+          "type": "string",
+          "description": "Markdown-formatted description content."
+        }
+      },
+      "minProperties": 1,
+      "example": {
+        "plain": "Classic cotton tee in red, size small."
+      }
+    },
+    "Price": {
```
prateek-ct, line 29: Thanks for this SEP, the product feed design is clean and the variant-centric model is the right call.                               
                                                                                                                                        
 One gap worth addressing before this lands-  the Price schema has no way to communicate whether amount is tax-inclusive or tax-exclusive. 

This is a critical ambiguity for global commerce, without tax_behaviour, an agent receiving `{ "amount": 1999, "currency": "EUR" } ` cannot know whether €19.99 includes VAT (true     
  price ~€16.66 pre-tax) or whether tax will be added on top. 

This matters for:
  - Price display: Consumer protection laws in many countries require showing tax-inclusive prices.                                     
  - Checkout reconciliation: When a feed item flows into a LineItem in a checkout session, the checkout totals need to know if tax has already been factored into unit_amount.                                                                                             
                                                                                                                                        
**Proposed addition to Price:**
                                                                                                                                        
```
  "tax_behaviour": {
    "type": "string",                                                                                                                   
    "enum": ["inclusive", "exclusive"],
    "description": "Whether the amount includes tax (inclusive) or excludes it (exclusive). Sellers SHOULD always set this field. When omitted, behaviour is seller-defined and agents SHOULD treat it tax exclusive."                                     
  }
```
                                                                                                                                        
Not a blocker to the SEP and this PR. Happy to submit a follow-up PR adding this if the design direction is agreed.

spec/unreleased/json-schema/schema.feed.json
```diff
@@ -0,0 +1,626 @@
...
+        },
+        "list_price": {
+          "$ref": "#/$defs/Price",
+          "description": "Reference or pre-discount price for the variant."
+        },
+        "unit_price": {
+          "$ref": "#/$defs/UnitPrice",
+          "description": "Normalized unit price, when applicable."
+        },
+        "availability": {
```
butenkor, line 347: Did we consider multiple prices and availabilities? For example a bike store could have many OTC stores with their own availabilities. I could let LLM to list me top 3 closest retailers with desired bike and on stock.

Also, availabilities/prices have their own much more frequent change occurrence than product. Stock and price details are often decoupled from the product data. An alternative could be a soft link from price/stock to the variant by unique variant ID. Same mechanic could be applied also to the variants which could live outside of the product and be linked to it over the soft link/id. This would omit race conditions (there is an availability but no product yet), allow smaller create/update payloads, allow much more variants associated with a single product. In addition, later on one could introduce channelled (region/customer group specific) prices/availabilities.

lhwa:
I have a few clarifying questions about the intended use of this Feed API:
1. **Offline ingestion vs. real-time fetching**: When is this API expected to be called relative to the checkout flow (offline or real time before checkout)?
2. **Single-product lookup vs. full-feed fetch**: The current API only supports retrieving the entire product set via GET /feeds/{id}/products. There isn’t an endpoint for fetching a single product by ID (e.g., GET /feeds/{id}/products/{product_id}). If an agent already knows which product the buyer wants (e.g., from prior search/recommendation or crawling the seller’s site), we would need single product lookup. Is the intent that agents always fetch the full feed, or should single-product retrieval be supported?
3. **Agent pull API vs. merchant push feed**: As implemented in this PR, this appears to be a pull model (the agent calls GET /feeds/{id}/products to fetch products from the seller). This is somewhat different from the traditional product feed approach where the merchant pushes a feed file to the platform. Can you confirm this is the intended model?

dylanfarrell:
Thank you for the SEP - this looks great.

One potential addition is treating FAQs as a first-class concept. A significant portion of high-value commerce content is expressed in question form, and preserving that structure could help improve consumer confidence as well as model performance by making product context more explicit.

Curious if this was considered?

examples/unreleased/examples.feed.json
```diff
@@ -0,0 +1,107 @@
...
+    "target_country": "US",
+    "updated_at": "2026-03-01T00:00:00Z"
+  },
+  "get_feed_response": {
+    "id": "feed_8f3K2x",
+    "target_country": "US",
+    "updated_at": "2026-03-01T00:00:00Z"
+  },
+  "get_products_response": {
+    "products": [
```
lhwa, line 16: If this is intended to return the full product set, we will need the product count and pagination.

changelog/unreleased/add-feed-api.md
```diff
@@ -0,0 +1,21 @@
+# Feed API
+
+**Added** – an unreleased Feed API surface for feed metadata and product catalog management.
+
+## New API Surface
```
prasad-stripe, line 5: Can we add more details like, these endpoints will be hosted by Agent and called by merchant, making it clear that this is a push model, merchants will push the product catalog to Agents.

prasad-stripe:
One thing worth flagging (already flagged): GET /feeds/{id}/products currently returns the full product set in a single response with no pagination, and PATCH /feeds/{id}/products has no batch size limit on the upsert payload. For merchants with large catalogs (tens of  thousands to millions of products), this could be a real problem in practice.

Also, the initial bulk load story after POST /feeds could use some thought — if a merchant creates a feed and wants to populate it with their full catalog, the only API path is repeated PATCH calls with no guidance on batch sizes. The products.jsonl file ingestion is mentioned in the description but isn't spec'd as an endpoint, so it's unclear how implementers should handle that path.

Some things that would be good to think through (totally fine as a fast follow-up PR):
  - Cursor-based or offset pagination on GET /feeds/{id}/products
  - maxItems or recommended batch size on UpsertProductsRequest.products
  - Guidance on max payload sizes
  - A defined protocol for the bulk file ingestion path

Not a blocker for this PR — just wanted to call it out so we can track it as a follow-up!

spec/unreleased/json-schema/schema.feed.json
```diff
@@ -0,0 +1,626 @@
...
+    },
+    "Variant": {
+      "description": "Purchasable variant of a product within a feed.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["id", "title"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "Stable global identifier for this variant."
```
lhwa, line 313: The variant id field is an important field in the feed schema since this is what is used to indicate purchase intent during checkout. Can we make the description more explicit about its uniqueness and intended source/usage?

1. Uniqueness: the id must be unique across the entire feed.  
Cross referencing the UCP definition 
"id": {
      "type": "string",
      "description": "Global ID (GID) uniquely identifying this variant. Used as item.id in checkout."
    }" 

2. Clarify relationship to SKU 
- UCP explicitly separated id from sku.  Ref: https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/variant.json
- Meta catalog API defines the id field as "A unique content ID for the item. Use the item's SKU if possible. Each content ID must appear only once in your catalog."  Ref: https://developers.facebook.com/documentation/ads-commerce/commerce-platform/catalog/fields#supported-feed-formats

changelog/unreleased/add-feed-api.md
aravindrao-openai: Done!

prasad-stripe (approved):

PR #249, opened 2026-05-12
fix(spec/2026-01-30): refund amount example type matches schema

Ectsang:
## Summary

The `order_updated` example in `spec/2026-01-30/openapi/openapi.agentic_checkout_webhook.yaml` showed `amount: "1.00"` (string) for a refund. The `Refund.amount` schema in the same file (lines 111-114) defines `amount` as an `integer` in minor units (e.g. `100` cents for `$1.00`). The example contradicted its own schema.

Other examples in the file (e.g. line item `total` at line 117) correctly use integer minor units.

Fixes #220.

## Changes

- `spec/2026-01-30/openapi/openapi.agentic_checkout_webhook.yaml` line 68: `amount: "1.00"` → `amount: 100`
- `changelog/unreleased/fix-refund-amount-example-type.md` added

## Scope

Single-version fix per the issue's citation of 2026-01-30. The newer `2026-04-17` spec replaces `refunds[]` with `adjustments[]` (already integer), so no fix needed there. Older spec versions (`2025-09-29`, `2025-12-12`, `2026-01-16`) contain the same string-form example — happy to extend the PR if maintainers prefer those frozen releases be left as-is or also corrected.

## Checklist

- [x] CLA — will sign on first PR via the bot
- [x] Branch from main: `fix/refund-amount-example-type`
- [x] Changelog entry added in `changelog/unreleased/`
- [x] Single-purpose, minimal scope
- [x] Linked issue: #220

PR #243, opened 2026-05-02
fix: unwrap error envelope in delegate_payment OpenAPI examples

sakinaroufid:
## Type of Change

- [ ] Documentation fix/improvement
- [x] Bug fix (non-breaking)
- [ ] Tooling improvement
- [x] Example update
- [ ] Minor data/enum addition
- [ ] Other:

---

## Description

The error response examples in `spec/unreleased/openapi/openapi.delegate_payment.yaml` wrap the body in `{ "error": { ... } }`. The `Error` schema is `additionalProperties: false` and requires `type`, `code`, `message` at the top level. RFC `rfc.agentic_checkout.md` §3 documents the same flat shape.

0 of the 9 error response examples validate against the `Error` schema they reference.

This PR removes the `error:` wrapper from all 9 examples (400, 401, 409, 422, 429, 500, 503). No `type`, `code`, or `message` values are changed.

---

## Motivation and Context

Same shape mismatch was raised in #21 (Oct 2025), which targeted the pre-versioned `spec/openapi/...` layout and is no longer mergeable. The `Error.code` enum gaps for 401/500/503 are out of scope here and being handled in #161. Leaving `type`/`code` values alone keeps this PR independent of #161 either way it lands.

Fixes part of: #21

---

## Testing

Validated all 9 examples against `components.schemas.Error` with AJV 8 (the version pinned in `package.json`):

| Status | Example | Before | After |
|---|---|---|---|
| 400 | invalid_request_missing_field | fail | pass |
| 400 | idempotency_key_required | fail | pass |
| 401 | unauthorized | fail | enum-only (#161) |
| 409 | idempotency_in_flight | fail | pass |
| 422 | invalid_request | fail | pass |
| 422 | idempotency_conflict | fail | pass |
| 429 | rate_limit_exceeded | fail | enum-only (#161) |
| 500 | internal_server_error | fail | enum-only (#161) |
| 503 | service_unavailable | fail | enum-only (#161) |

0/9 pass before, 5/9 pass after. The 4 "enum-only" rows are structurally correct but use values outside the current `type`/`code` enums.

`pnpm run compile:schema` and `pnpm run validate:all` both pass locally.

---

## Screenshots / Examples

Before:
```yaml
"401":
  content:
    application/json:
      schema:
        $ref: "#/components/schemas/Error"
      examples:
        unauthorized:
          value:
            error:
              type: unauthorized
              code: unauthorized
              message: Unauthorized
```

After:
```yaml
"401":
  content:
    application/json:
[...]

PR #161, opened 2026-02-24
Delegate Payment: extend Error type and code enums for 401/500/503

Kixo:
## 🔧 Type of Change

- [ ] Documentation fix/improvement
- [x] Bug fix (non-breaking)
- [ ] Tooling improvement
- [x] Example update
- [x] Minor data/enum addition
- [ ] Other: [describe]

---

## 📝 Description

Extends the delegate payment **Error** schema so all documented HTTP error responses validate:

- **Error.type** enum: added `unauthorized` (401) and `internal_server_error` (500).
- **Error.code** enum: added `unauthorized`, `internal_server_error`, and `service_unavailable` (503).

Previously, the OpenAPI spec documented 401, 500, and 503 responses with examples that used these type/code values, but the schema did not allow them, so the examples failed validation.

**Forward compatibility (RFC):** Added a short "Forward compatibility for error codes" subsection to **§4.2 Error** in `rfcs/rfc.delegate_payment.md`. The `type` and `code` enums are defined as extensible; implementations MUST NOT reject unrecognized values and SHOULD fall back to generic error handling; strict enum validation of error responses is NOT RECOMMENDED. This protects future additions (e.g. x402 or auth codes) and aligns with intent traces (§7.2) and capability negotiation (§6.4, §7.3).

**Files changed:**

- `spec/unreleased/json-schema/schema.delegate_payment.json` — Error type and code enums
- `spec/unreleased/openapi/openapi.delegate_payment.yaml` — Error type and code enums
- `examples/unreleased/examples.delegate_payment.json` — added error examples for unauthorized, internal_server_error, service_unavailable
- `rfcs/rfc.delegate_payment.md` — added **Forward compatibility for error codes** under §4.2
- `changelog/unreleased/delegate-payment-error-enums.md` — changelog entry (includes enum extension + RFC forward-compat)

---

## 🎯 Motivation and Context

- **Schema vs examples:** The spec had `type` enum: `invalid_request`, `rate_limit_exceeded`, `processing_error`, `service_unavailable` and `code` enum without `unauthorized`, `internal_server_error`, or `service_unavailable`. The OpenAPI examples for 401 (Unauthorized), 500 (Processing error), and 503 (Service unavailable) used `type: unauthorized` / `code: unauthorized`, `type: internal_server_error` / `code: internal_server_error`, and `code: service_unavailable`, so 401 and 500 were invalid against the schema; 503 had valid type but code was missing from the enum.
[...]

PR #21, opened 2025-10-06
fix: resolves consistency issues in spec and examples

ericlakich:
# Agentic Checkout Spec Review

---

## ~Issue-1: Type Mismatch in *FulfillmentOption* Fields~

> [!NOTE]
> This issue is resolved by PR #11  

**Location:** `spec/openapi/openapi.agentic_checkout.yaml:380–395`

**Problem:**
In the `FulfillmentOptionShipping` and `FulfillmentOptionDigital` schemas, the `subtotal`, `tax`, and `total` fields are defined as `type: string`, but they should be `type: number` to match:

1. The JSON Schema definition (`schema.agentic_checkout.json:183–220`)
2. The RFC documentation, which explicitly states: “All money fields are integers (minor units)”
3. The example data, which uses integer values (e.g., `"subtotal": 100`)

**Impact:**
This inconsistency could lead to integration errors where implementations following the OpenAPI spec send strings while the JSON schema and examples expect integers.

---

## Issue-2: Missing `payment_provider` in Response Examples

**Location:** `examples/examples.agentic_checkout.json`

**Problem:**
Several response examples are missing the `payment_provider` field, which is **required** in the `CheckoutSessionBase` schema. Specifically:

* `update_checkout_session_response` (line 115)
* `get_checkout_session_response` (line 320)
* `cancel_checkout_session_response` (line 409)

These responses should include the `payment_provider` object to be valid according to the schema.

---

## Issue-3: Missing `order` Field in Complete Response Example

**Location:** `examples/examples.agentic_checkout.json:225` (`complete_checkout_session_response`)

**Problem:**
The `complete_checkout_session_response` example is missing the `order` field, which is required by the `CheckoutSessionWithOrder` schema and the RFC specification (Section 4.4: “Response MUST include status: completed and an order”).

---

## Issue-4: Inconsistent Error Response Format in Delegate Payment API

**Location:** `spec/openapi/openapi.delegate_payment.yaml:105–121`

**Problem:**
The error examples in the Delegate Payment API wrap the error object in an `error` property, e.g.:

```json
{
  "error": { "type": "...", "code": "...", ... }
}
```

…but the schema definition at lines 409–430 shows the `Error` object is returned **directly** (not wrapped).

**Impact:**
This inconsistency could cause confusion for implementers.
The schema should match the examples—or vice versa.

---

## Issue-5: Invalid Error Codes in Delegate Payment Examples

**Location:** `spec/openapi/openapi.delegate_payment.yaml`

**Problem:**
[...]

PR #252, opened 2026-05-16
ci: validate OpenAPI inline examples against $ref schemas

sakinaroufid:
## 🔧 Type of Change
- [x] Tooling improvement

## 📝 Description
This adds a CI check and nothing else. Following the review note, I moved the example fix out into #266, so this PR is just the validator.

`scripts/validate-consistency.js` already walks `examples/<version>/examples.<spec>.json`, but it never looked at the inline `examples:` embedded in the OpenAPI `requestBody` and `responses` blocks. That gap is how the wrapped error-envelope bug fixed in #243 shipped across several versions without CI noticing.

The new `validateOpenApiInlineExamples()` walks each `content[mediaType]` block, resolves `schema.$ref` against the document's `#/components/schemas/*`, and validates every inline `example` and `examples[*].value` against it with AJV. OpenAPI 3.0 conventions (boolean `exclusiveMinimum`/`exclusiveMaximum` and `nullable: true`) are normalized to JSON Schema 2020-12 before compiling. Cross-file `$ref`s and inline schemas without a `$ref` are skipped, so it does not raise false positives.

The pre-existing failures it surfaces are listed in `KNOWN_INLINE_EXAMPLE_FAILURES` so the check can land green. Each entry is a real example bug for its own follow-up. The released `delegate_payment` envelope examples are unwrapped in #266, and the six allowlist entries covering them come out once that lands.

## 🎯 Motivation and Context
A consistency check that ignores inline OpenAPI examples lets schema and example drift slip through, which is exactly what happened with the error envelope. Landing the check on its own, with existing failures tracked in an allowlist, keeps it from dragging a pile of example edits along with it.

## 🧪 Testing
`pnpm run compile:schema` and `pnpm run validate:all` both pass with the allowlist in place. I also confirmed the check goes red on the wrapped examples when the allowlist entries are removed, so it catches the bug it is meant to catch.

## 📸 Screenshots / Examples
Check output when the envelope examples are not allowlisted:

```
❌ ERROR: OpenAPI inline example "invalid_request_missing_field" at
   paths['/agentic_commerce/delegate_payment'].post.responses.400.content[application/json]
   does not validate against #/components/schemas/Error
```

## ✅ Checklist
- [x] My change follows the project's code style and conventions
- [x] I updated docs (changelog entry; header comment in `validate-consistency.js`)
- [x] I added a changelog entry at `changelog/unreleased/validate-openapi-inline-examples.md`
- [x] I tested locally (`pnpm run compile:schema` and `pnpm run validate:all`)
- [x] This change does not require a SEP (tooling only)

## 🔍 Scope Verification
- [x] Minor, straightforward change. It adds a CI check and an allowlist. No protocol, schema, or example values change here.

## 📚 Additional Notes
Companion PR: #266 unwraps the released `delegate_payment` error examples and removes the six allowlist entries for them.

The remaining allowlist entries cover separate pre-existing example bugs (items vs line_items rename, missing required fields, webhook refund amount typed as string, extra fields in delegate_authentication examples, and the Error enum gaps in #161). Each can be removed by a focused follow-up.

Tracks #243, #249, #161, #21.

### DIFF
--- a/scripts/validate-consistency.js
+++ b/scripts/validate-consistency.js
@@ -13,6 +13,7 @@
  * 7. All data models in unreleased JSON schemas have at least one example
  * 8. All fields in unreleased OpenAPI schemas have descriptions
  * 9. All schemas in unreleased OpenAPI specs have at least one example
+ * 10. Inline examples under requestBody/responses validate against the schema they $ref
  */
 
 const fs = require('fs');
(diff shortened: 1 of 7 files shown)


scripts/validate-consistency.js
lhwa, line 1: Can we split the validator change into a separate PR from the inline example fixes?

lhwa (commented):
This PR bundles two changes: (1) adding new CI validator logic for the consistency checks, and (2) fixing the error envelope in the inline examples. Should we split the validator changes into a separate PR?

sakinaroufid:
@lhwa Thanks for the review. I split it into two PRs. This one (#252) is now just the validator and the allowlist. The released `delegate_payment` envelope unwrap moved to #266. Both pass CI on their own, and I checked that the validator stays green against the unwrapped examples once both land, so they can go in in either order.

sakinaroufid:
@lhwa Following up on this!

Issue #280, opened 2026-07-08
SEP: Add optional product page URL to checkout Item


vig-star:
## Summary

Add an optional `url` (product page URL) field to the request-side `Item` object used in `line_items` on checkout session create/update. This gives sellers a canonical, standard signal to resolve exactly which product a line item refers to when a single seller account spans multiple brands or storefronts that cross-list the same item identifier or SKU.

## Motivation

The request-side `Item` today carries only `id` (required), `name`, `unit_amount`, and `suggested_price`. None of these reliably identify *which* product/catalog entry a line item maps to when a seller operates multiple brands or storefronts under one account and the same `id` or SKU is listed under more than one of them.

In that situation the seller has no deterministic way, at checkout, to tell which brand's product is being purchased — which can lead to incorrect catalog resolution, pricing, tax, or fulfillment.

A product page URL is a natural, already-available artifact on the agent side (the page the item was sourced from) and uniquely identifies the exact product and brand, with no new taxonomy or registry. Notably, a `url` ("Product page URL") already exists on the post-checkout `OrderLineItem`, but there's no equivalent on the request-side `Item`, so the agent can't communicate it where disambiguation is actually needed.

## Proposed change

Add an optional `url` to `$defs.Item` (and its OpenAPI mirror):

```json
"url": {
  "type": "string",
  "format": "uri",
  "description": "Product page URL for this item. When present, sellers MAY use it to resolve the specific product the line item refers to (for example, to disambiguate when the same item identifier or SKU is cross-listed across multiple brands or storefronts under a single seller account)."
}
```

Optional (not added to `required`); reuses the exact name/type/`format: uri` of the existing `OrderLineItem.url`.

## Backward compatibility

Fully additive. The field is optional; existing agents and sellers are unaffected. (Because `Item` sets `additionalProperties: false`, it must be defined in the schema for agents to send it — it can't be passed today.)

## Alternatives considered

- **A dedicated `brand`/`seller_brand` identifier** — would require a shared brand taxonomy/registry and is merchant-specific; a URL is already available and canonical.
- **Rely on `sku`/`variant_id`** — these are exactly the identifiers that are ambiguous here, and they live on the response `LineItem`, not the request `Item`.
- **Use `metadata`** — unstructured and merchant-specific; defeats the cross-agent/cross-seller interoperability that's the whole point.
- **Name it `product_url`** — rejected in favor of `url` to match the existing `OrderLineItem.url` precedent.

[...]

PR #201, opened 2026-03-26
SEP: Suggested Pricing on Checkout Create

lhwa:
# SEP: Suggested Pricing on Checkout Create

## 📋 SEP Metadata

- **SEP Number**: [[#197](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/197)](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/197)
- **Author(s)**: Lee Hwa 
- **Status**: `proposal` (awaiting sponsor)
- **Type**: [x] Major Change [ ] Process Change
- **Related Issues/RFCs**: rfcs/rfc.agentic_checkout.md, spec/2026-01-30/json-schema/schema.agentic_checkout.json

---

## 🎯 Abstract

This SEP proposes adding an optional `suggested_price` field to `Item` on `CheckoutSessionCreateRequest`. Agent platforms maintain product catalogs synced from sellers, but pricing data inevitably drifts — a buyer may discover a product at a sale price that has since expired. Currently, the agent has no way to communicate price provenance to the seller. By including `suggested_price` with a structured `PriceSource` (feed identity, update version, timestamp, and channel) and an observation timestamp, the agent provides the seller with the information needed to make an informed decision about whether to honor the price the buyer saw. This is a best-effort mechanism — if the seller does not honor the suggested price, the checkout session proceeds normally with the seller's current pricing breakdown. No error is raised.

---

### The Problem

Agent platforms display product pricing sourced from seller catalog feeds, product search API, or web crawling. This catalog data is refreshed periodically — every few minutes, hours, or daily depending on the seller and integration. Between refreshes, seller-side pricing can change due to:

- Flash sales ending or starting
- Dynamic pricing adjustments
- Inventory-based price changes
- Promotional period expirations

When a buyer discovers a product at one price on the agent platform and initiates checkout, the seller may return a different price. The buyer experiences an unexpected price change at the most sensitive moment — after they've decided to purchase.

### Why the Community Should Care

- **Buyers** lose trust when prices change between discovery and checkout. Price increases at checkout are a leading cause of cart abandonment.
- **Agent developers** have no protocol-level mechanism to communicate the price they displayed to the buyer. Every agent must implement ad-hoc workarounds (metadata fields, custom headers, out-of-band agreements with sellers).
- **Sellers** want to honor prices from their own catalog feeds — they published those prices intentionally. But without knowing which feed or when the agent observed the price, they can't distinguish a legitimate catalog price from an arbitrary number.
[...]

Issue #197, opened 2026-03-25
Issue: Suggested Pricing on Checkout Create


lhwa:
# Issue: Suggested Pricing on Checkout Create

**Labels:** `SEP`, `proposal`

## Title

Add suggested_price to Item on CheckoutSessionCreateRequest

### Problem

Agent platforms maintain product catalogs that are synchronized from sellers or business solution providers. These catalogs include pricing data — regular prices and sale prices — that the agent displays to buyers during product discovery. However, by the time the buyer initiates a checkout, the seller's live pricing may have changed. The problem of agent catalog data going stale is inevitable, no matter how frequently it is synced.

This creates a pricing discrepancy: the buyer discovers a product on the agent platform at one price (e.g., a sale price), but when the agent creates a checkout session, the seller returns a different price (e.g., the regular price because the sale ended). The buyer sees an unexpected price increase at checkout — one of the most trust-damaging experiences in commerce.

### Why This Is a Problem

The current `CheckoutSessionCreateRequest` sends `Item` objects with an optional `unit_amount`, but this field carries no provenance — no indication of where the price came from or when it was observed. The seller has no way to know:

- Whether the agent displayed a specific price to the buyer
- Where the agent obtained that price (which catalog feed, API, or snapshot)
- When the agent last observed that price

Without this context, the seller cannot make an informed decision about whether to honor the agent's displayed price. The seller simply returns its current live price, which may differ from what the buyer saw.

### Current Spec Behavior

- `CheckoutSessionCreateRequest.line_items`: array of `Item` objects ([schema.agentic_checkout.json](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/blob/main/spec/2026-01-30/json-schema/schema.agentic_checkout.json))
- `Item`: has `id` (required), `name` (optional), `unit_amount` (optional)
- `unit_amount`: simple integer with no source or timestamp — the seller cannot distinguish between an agent-provided price and a stale catalog price
- No mechanism exists for the agent to communicate price expectations or provenance to the seller

### Proposed Solution

Add an optional `suggested_price` field to `Item` on `CheckoutSessionCreateRequest`. This object includes:

- **`amount`**: The price per unit the agent displayed to the buyer, in minor currency units
- **`source`**: A URI or identifier for the catalog source where the price was obtained (e.g., a product feed URL, catalog API endpoint, or snapshot ID)
[...]

PR #281, opened 2026-07-08
SEP: Optional product URL on checkout Item

vig-star:
# SEP: Optional product URL on checkout Item

## 📋 SEP Metadata

- **SEP Number**: #280
- **Author(s)**: Vignesh Sreedhar, Meta
- **Related Issues/RFCs**: SEP issue #280; precedent `OrderLineItem.url`; SEP #197 (`suggested_price` on `Item`)

---

## 🎯 Abstract

The request-side `Item` object (used in `line_items` on checkout session create and update) currently identifies a purchasable item with `id`, `name`, `unit_amount`, and `suggested_price`. It has no field tying the item to a specific product page. When a single seller account spans multiple brands or storefronts, and the same item identifier or SKU is cross-listed across more than one of them, the seller has no deterministic way at checkout to resolve which product — and therefore which brand — a line item refers to. This ambiguity can produce incorrect catalog resolution, pricing, tax, or fulfillment. This SEP adds an optional `url` field (a canonical product page URL, `type: string`, `format: uri`) to `Item`. When present, sellers can map the URL to the exact product and brand. The field reuses the identical name, type, and format already defined on the post-checkout `OrderLineItem.url`, introducing no new conventions. It is optional and purely additive, so existing agents and sellers are unaffected. Because `Item` declares `additionalProperties: false`, the field must be defined in the schema for agents to send it — this change is what makes the signal available in a standard, interoperable way.

---

## 💡 Motivation

Some sellers operate multiple brands or storefronts under a single seller account. The same physical product — and sometimes the same `id` or SKU — is cross-listed across more than one of these brands. At checkout, the request-side `Item` provides no signal that distinguishes them, so the seller cannot deterministically resolve which brand/catalog entry a line item belongs to. The consequences are concrete: wrong catalog match, wrong price, wrong tax treatment, wrong fulfillment path.

The protocol already recognizes a product page URL as meaningful item-identity data — `OrderLineItem` carries `url` ("Product page URL") post-checkout. The gap is that this signal is unavailable at the point of checkout, on the request-side `Item`, where the disambiguation decision is made.

---

## 📐 Specification

This change is confined to `spec/unreleased/` and follows the same four-file shape as SEP #197 (PR #201):

1. `spec/unreleased/json-schema/schema.agentic_checkout.json` — add `url` to `$defs.Item.properties` and to `$defs.Item.example`.
2. `spec/unreleased/openapi/openapi.agentic_checkout.yaml` — mirror the addition on `components/schemas/Item` (CI enforces schema/OpenAPI parity).
3. `examples/unreleased/examples.agentic_checkout.json` — include `url` on a line item in the create-request example.
4. `changelog/unreleased/product-url-on-checkout-item.md` — new changelog entry.

Field definition (JSON Schema; OpenAPI mirror is equivalent):

```json
"url": {
  "type": "string",
  "format": "uri",
  "description": "Product page URL for this item. When present, sellers MAY use it to resolve the specific product the line item refers to (for example, to disambiguate when the same item identifier or SKU is cross-listed across multiple brands or storefronts under a single seller account)."
}
```

`url` is NOT added to `Item.required`. No other objects change.

---

## 🤔 Rationale

- **Why a URL, not a new `brand` field?** A brand/seller identifier would require a shared taxonomy or registry and is inherently merchant-specific. A product page URL is already available to the agent, is canonical, and needs no new vocabulary.
- **Why not reuse `sku`/`variant_id`?** Those are precisely the identifiers that are shared/ambiguous in this scenario, and they exist on the response `LineItem`, not the request `Item`.
- **Why not `metadata`?** The request already allows arbitrary `metadata`, but an unstructured, merchant-specific bag defeats cross-agent/cross-seller interoperability. A named, standard field is what makes the signal portable.
- **Why `url` and not `product_url`?** To match the existing `OrderLineItem.url` for the same concept; the spec uses bare `url` on product-like objects (`OrderLineItem`, `Link`, `AffiliateAttributionSource`).
- **Why request-side only?** Disambiguation happens when the agent submits the cart; the seller needs it as input. A response-side echo can be added later if desired.

---

## 🔄 Backward Compatibility

No breaking changes. The field is optional and additive; existing agents and sellers continue to work unchanged. Because `Item` uses `additionalProperties: false`, this addition is what enables agents to send the field at all — there is no prior behavior to preserve or migrate.

---

Available on request; the change is a schema/OpenAPI/example/changelog addition with no runtime logic required by the protocol itself.

---

## 🔒 Security Implications

The `url` is supplied by the agent and consumed by the seller. Sellers SHOULD treat it as an untrusted hint/identifier — preferably matching it against their own known catalog URLs (e.g., allowlisting their own domains) rather than dereferencing arbitrary URLs server-side. If a seller does fetch the URL, standard SSRF protections apply (block internal/link-local ranges, disallow redirects to internal hosts, apply timeouts). The field conveys a public product page URL and introduces no new PII or sensitive data. Because `Item` uses `additionalProperties: false`, `url` is an explicit, controlled schema addition rather than an open-ended field.

---

## ✅ Pre-Submission Checklist

- [x] This PR includes a changelog entry file in `changelog/unreleased/`
- [x] I am seeking or have found a sponsor

---

## 📚 Additional Context

- Precedent for shape and file-set: SEP #197 / PR #201 ("Suggested Pricing on Checkout Create"), which added an optional field to `Item`.
- Precedent for the field itself: `OrderLineItem.url` ("Product page URL", `format: uri`).

---

## 🙋 Questions for Reviewers

1. Preference on echoing `url` on the response `LineItem` now vs. leaving it as a future additive extension?
2. Confirm `url` (matching `OrderLineItem.url`) is preferred over `product_url`.


aravindrao-openai (approved):

aravindrao-openai:
Approved
1. I think this is a valid change given there is a real world use case for this
2. `url` is better since it matches the same field name in the product feed

PR #296, opened 2026-08-13
Require Idempotency-Key for mutating MCP checkout tools

### LINKED ISSUE #295: MCP binding makes the mandatory `Idempotency-Key` optional for mutating checkout tools
## Description

Agentic Checkout requires clients to send an `Idempotency-Key` on every POST request, including create, update, complete, and cancel. The MCP binding maps those operations to tools but marks `meta.idempotency_key` as not required and recommends it only for create and complete.

The OpenRPC `meta.required` array contains only `api_version`, even though the binding says that MCP preserves REST idempotency with identical semantics. The canonical update and cancel examples also omit the key. See `rfcs/rfc.agentic_checkout.md:232-249`, `docs/mcp-binding.md:46-56,102-116,255-263`, `spec/unreleased/openrpc/openrpc.agentic_checkout.json:21-36`, and `examples/unreleased/examples.mcp.agentic_checkout.json:213-237,352-371`.

This conflicts with the mandatory idempotency contract introduced by [SEP #120](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/120) and merged in [PR #121](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/121). The MCP binding was introduced by [SEP #135](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/issues/135) and [PR #139](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol/pull/139), but its schema still permits the older optional behavior.

## Impact

- MCP-generated clients may omit the key from state-changing checkout calls.
- After a timeout or lost response, retrying `complete_checkout_session` may repeat payment capture or order creation instead of returning the original result.
- Update and cancel tools also have weaker retry semantics than their REST equivalents.
- This is a specification and code-generation risk; actual production impact depends on whether the MCP server independently rejects missing keys or derives and reuses a stable key across retries.

## Reproduction Steps

1. Review `rfcs/rfc.agentic_checkout.md:232-249` and confirm that every POST request MUST include `Idempotency-Key` and that a missing key MUST be rejected with HTTP 400.
2. Review `docs/mcp-binding.md:50-56` and confirm that create, update, complete, and cancel map directly to those POST operations.
3. Review `docs/mcp-binding.md:107-115` and observe that `meta.idempotency_key` is marked `Required: No`.
4. Review `spec/unreleased/openrpc/openrpc.agentic_checkout.json:21-36` and confirm that `meta.required` contains only `api_version`.
5. Review `examples/unreleased/examples.mcp.agentic_checkout.json:213-237,352-371`. The canonical update and cancel requests omit `idempotency_key`; the OpenRPC `meta` schema accepts those shapes, whereas the equivalent REST requests must be rejected.
6. As a safe control, define a mutating-tool `meta` schema that requires `idempotency_key` for create, update, complete, and cancel while keeping it optional for `get_checkout_session`. Require the MCP server to preserve the key and REST replay behavior when forwarding the operation.

ShlokShar:
Fixes #295

Align the MCP binding with the checkout idempotency contract by requiring `meta.idempotency_key` for create, update, complete, and cancel operations. Also update the binding docs and canonical examples to show the required key on mutating tool calls.


shunhe-wang (changes requested):
Hi all, suggesting a follow on. The mutation requirement is correct, but putting `idempotency_key` in the shared `#/components/schemas/meta.required` also makes it mandatory for `get_checkout_session`, because all five methods reference that same component. That conflicts with the GET mapping and with the safe control described in #295.

Recommend splitting the component into read and mutation metadata schemas, or compose the requirement at each mutating method parameter. The prose table should also express that the key is conditionally required for create/update/complete/cancel rather than globally required.

A small schema-level regression matrix would make this durable:

1. create without `idempotency_key` -> reject;
2. update/complete/cancel without it -> reject;
3. GET without it -> accept;
4. replay a mutation with the same key and equivalent body -> preserve the original result;
5. reuse the key with a different body -> conflict.

The PR also needs the changelog entry required by CONTRIBUTING.md. The current patch fixes the four mutation examples, but without the GET control it introduces the inverse conformance error.

